PR #142, opened 2017-06-20
Cleaned up Coreutils and changed ref

jemten:


lib/MIP/Gnu/Coreutils.pm
```diff
@@ -0,0 +1,1013 @@
...
+    ## gnu_split
+    # Stores commands depending on input parameters
+    my @commands = qw(gnu_split);
+
+    ## Options
+    if ($lines) {
+        push @commands, '--lines=' . $lines;
+    }
+
+    if ($numeric_suffixes) {
```
henrikstranneheim: It looks like the $numeric_suffixes are evaluated twice in gnu_split

henrikstranneheim: Ok, now I remember it is supposed to be like that, but could you remove one if($numeric_suffixes) and move the 
```
push @commands, '--suffix-length=' . $numeric_suffixes;
```
to L781.

mip.pl
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+!/usr/bin/env perl
```
henrikstranneheim: You need to add the '#' again

jemten: Strange. I thought I added that... It's part of my local file. Will update

lib/MIP/Gnu/Coreutils.pm
```diff
@@ -0,0 +1,1013 @@
...
+            default     => 0,
+            allow       => [ 0, 1 ],
+            strict_type => 1,
+            store       => \$verbose
+        },
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak qw[Could not parse arguments!];
+
+    my $SPACE = q{ };
```
henrikstranneheim: Let's start making the constants proper globals and just define them once per perl module

Add: 
```
## Constants
 my $SPACE = q{ };
```
etc after the begin block

PR #153, opened 2017-06-26
test for gnu_echo

jemten:


t/gnu_echo.t
```diff
@@ -0,0 +1,179 @@
...
+        inputs_ref      => [qw(string.test)],
+        expected_output => '"string.test"',
+    },
+);
+
+## Specific arguments
+my %specific_argument = (
+    strings_ref => {
+        inputs_ref      => [qw(string.test)],
+        expected_output => '"string.test"',
```
henrikstranneheim: Wil it look exactly like '"string.test"' or can you write it like q{"string.test"} or even
```
Readonly my $DOUBLE_QUOTE => q{"};
$DOUBLE_QUOTE . q{string.test} . $DOUBLE_QUOTE
```

jemten: This will work: 
```Perl
strings_ref => {
    inputs_ref => [q{This is my test string}],
    expected_output => $DOUBLE_QUOTE . q{This is my test string} . $DOUBLE_QUOTE,
}
```
When running gnu_echo with that should give 
```Bash
echo "My test string"
```
I'll upload the changes

PR #154, opened 2017-06-26
test for gnu_printf and minor corection of gnu_printf

jemten:
print to stdoutfile_path was not implemented in gnu_printf


henrikstranneheim:
👍

PR #169, opened 2017-08-24
Sambamba into lib/MIP/Program/Alignment

northwestwitch:
Fixed and moved Sambamba to lib/MIP/Program/Alignment. Changed mip.pl accordingly. Created subroutine tests under t/sambamba. No problems wih pull request this time!


mip.pl
```diff
@@ -41,9 +41,7 @@
 use File::Format::Yaml qw(load_yaml write_yaml);
 use MIP_log::Log4perl qw(initiate_logger);
 use Script::Utils qw(help);
-use MIP::File::Format::Pedigree qw(create_fam_file);
```
henrikstranneheim: It is strange that these lines are missing from the mip.pl script. They should be there. There is still something strange with this I cannot merge this

PR #172, opened 2017-08-28
Feature/fix undef conda env

jemten:
Only write instructions to mip.sh to activate conda environment if that parameter is defined


henrikstranneheim:
👍

PR #189, opened 2017-09-14
Feature/snpeff download

jemten:
The test might require some polishing


lib/MIP/Program/Variantcalling/SnpEff.pm
```diff
@@ -0,0 +1,161 @@
...
+    my $stderrfile_path;
+    my $stderrfile_path_append;
+    my $stdoutfile_path;
+
+    my $tmpl = {
+        FILEHANDLE => {
+            required => 1,
+            store    => \$FILEHANDLE,
+        },
+        memory_allocation => {
```
henrikstranneheim: Can you collect all the options with default values at the end of the sub doc header, $tmpl. They are the ones one usually care less about

t/snpeff_download.t
```diff
@@ -0,0 +1,218 @@
...
+##Modules
+    my @modules = (q{MIP::Program::Variantcalling::SnpEff});
+
+  MODULES:
+    for my $module (@modules) {
+        require_ok($module) or BAIL_OUT q{Cannot load} . $SPACE . $module;
+    }
+}
+
+use File::Spec::Functions qw{ catfile };
```
henrikstranneheim: Add this to the import with File:.Spec above instead

t/snpeff_download.t
```diff
@@ -0,0 +1,218 @@
...
+        memory_allocation => q{Xmx2g},
+        java_jar          => catfile(qw{ path to jar }),
+    }
+  ),
+  qw{ download };
+
+my %base_argument = (
+    stdoutfile_path => {
+        input           => q{stdoutfile.test},
+        expected_output => '1> stdoutfile.test',
```
henrikstranneheim: q{}

PR #201, opened 2017-09-27
Features/tiddit

northwestwitch:
Note that I couldn't test it because I think tiddit is switched off for wes analysis and I have wes samples for testing.


henrikstranneheim:
:1st_place_medal:

PR #205, opened 2017-10-01
pip module and test

jemten:


lib/MIP/Package_manager/Pip.pm
```diff
@@ -0,0 +1,137 @@
...
+    my $packages_ref;
+    my $quiet;
+    my $requirement;
+    my $editable;
+    my $stdoutfile_path;
+    my $stderrfile_path;
+    my $stderrfile_path_append;
+    my $FILEHANDLE;
+
+    ## Default(s)
```
henrikstranneheim: Remove

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+
+    # Set the version for version checking
+    our $VERSION = 1.0.0;
+
+    # Functions and variables which can be optionally exported
+    our @EXPORT_OK = qw{ install_pip_packages };
+
+}
+
+## Setup Conda env and install default packages
```
henrikstranneheim: One line down or?

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+
+sub install_pip_packages {
+
+## pip_install_pacakages
+
+## Function  : Recipe for install pip package(s).
+## Returns   :
+## Arguments : $pip_packages_href, $quiet, $conda_env, $FILEHANDLE
+##           : $conda_packages_href => Hash with pip packages and their version numbers {REF}
```
henrikstranneheim: $conda_packages_href does not look to be part of the rest of the template, I assume you mean $pip_packages_href

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+##           : $quiet               => Optionally turn on quiet output
+##           : $conda_env           => Name of conda environment
+##           : $FILEHANDLE          => Filehandle to write to
+
+    my ($arg_href) = @_;
+
+    ## Defaults
```
henrikstranneheim: Remove if not used

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+    my $tmpl = {
+        pip_packages_href => {
+            required    => 1,
+            defined     => 1,
+            default     => {},
+            strict_type => 1,
+            store       => \$pip_packages_href,
+        },
+        quiet => {
+            strict_type => 1,
```
henrikstranneheim: allow => [0, 1]??

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+        package_href => {
+            required    => 1,
+            defined     => 1,
+            default     => {},
+            strict_type => 1,
+            store       => \$package_href
+        },
+        package_version_separator => {
+            required    => 1,
+            defined     => 1,
```
henrikstranneheim: Could you set a default here? Perhaps q{==}?

henrikstranneheim: and then you wouldn't always have to supply the flag to the sub

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+            strict_type => 1,
+            store       => \$package_version_separator
+        },
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    my @packages;
+
+  PACKAGES:
```
henrikstranneheim: PACKAGES: -> PACKAGE:

lib/MIP/Recipes/Install/Pip.pm
```diff
@@ -0,0 +1,199 @@
...
+    while ( my ( $package, $package_version ) = each %{$package_href} ) {
+        if ( defined $package_version ) {
+
+            # Check that the version number matches pattern
+            if ( $package_version !~ qr/\d+.\d+ | \d+.\d+.\d+/xms ) {
+                croak q{The version number does not match defiend pattern for }
```
henrikstranneheim: Here we should add a LOGGER in the future

mip_install.pl
```diff
@@ -2162,9 +2077,15 @@ sub svdb {
     print $FILEHANDLE "\n\n";
 
     ## Install
-    print $FILEHANDLE '## Install', "\n";
-    print $FILEHANDLE 'pip install . ';
-    print $FILEHANDLE "\n\n";
+    say {$FILEHANDLE} q{## Install};
+    pip_install(
+        {
+            packages_ref => [qw{ . }],
```
henrikstranneheim: maybe [$DOT] is clearer and a comment

lib/MIP/Recipes/Install/Pip.pm
henrikstranneheim: why? Could it be 'undef' as well?

jemten: If the quiet flag is not supplied on the command line when running mip_install.pl, then the value is undefined

lib/MIP/Recipes/Install/Pip.pm
jemten: The plan was to make this function into a separate module since it's used in conda as well. However in conda the separator is a single '='. I could still have a default of course.

lib/MIP/Recipes/Install/Pip.pm
henrikstranneheim: [undef, 0, 1]

henrikstranneheim:
👍

PR #206, opened 2017-10-01
small changes to command module template

jemten:
Removed some redundant modules. These are not needed if the module is used as intended, in conjunction with recipe/master script).  Feel free to reject if you think that they are needed for other reasons :)


henrikstranneheim:
Looks great! 👍

PR #209, opened 2017-10-05
created Gatk.pm in Programs/Base, moved Gatk Variantcaller to under MIP::Program::Variantcaller

northwestwitch:
Changed mip.pl to call Gatk subroutines from new location.
Note that Gatk under Base is a bit complex so perlcritic complains.. 
Building Java commands in Gatk module now.
Tested on Rasta.


henrikstranneheim:
Oh no!

This was not what I intended. Only parameters shared in the common engine should go into base. The rest are specific and should reside within each command. Look at the picardtools example

Skickat från min iPad

northwestwitch:
It's ok, I have to fix other stuff as well, that's why I retired it.
I'll get back to you with the fixes shortly!


Chiara

On Thu, Oct 5, 2017 at 3:26 PM, Henrik Stranneheim <notifications@github.com

PR #221, opened 2017-10-29
Added color to MIP's screen log

jemten:
Suggestion to color the log's screen output for log messages with levels higher than info


henrikstranneheim:
Nice

PR #223, opened 2017-10-30
added cpan package and comments to bashrc/bash_profile

jemten:


perl_install.pl
```diff
@@ -297,6 +298,8 @@ sub install_perl {
     say {$FILEHANDLE} q{cd } . $pwd . qq{\n};
 
     ## Editing PATH and .bashrc
+    say {$FILEHANDLE}
+      q{echo '# Added by MIP's perl installer ' $(date) >> ~/.bashrc};
```
henrikstranneheim: Will this not cause $(date) to not be interpolated in bashrc as there is three ' in the string? Also it is always a good idea when writing bash to quote the variable i.e. "$(date)". Here is a nice help when writing bash: https://www.shellcheck.net/

henrikstranneheim:
👍

PR #235, opened 2017-11-08
small fix to snpeff_download

jemten:
discovered an error in snpeff_download while looking at fixing travis


henrikstranneheim:
👍

PR #238, opened 2017-11-10
Feature/vep flag fix

jemten:
Some of the array variables that could be set from the command line didn't function as intended before. Previously the argument that was given on the command line was appended to the array with the default values. Now the default array is overwritten instead


henrikstranneheim:
👍

PR #242, opened 2017-11-14
moved Peddy to lib/MIP/Programs/Variantcalling

northwestwitch:
Tested on rasta


lib/MIP/Program/Variantcalling/Peddy.pm
```diff
@@ -0,0 +1,163 @@
...
+        },
+        family_file_path => {
+            required    => 1,
+            defined     => 1,
+            strict_type => 1,
+            store       => \$family_file_path
+        },
+        processor_number => {
+            default     => 4,
+            allow       => qr/^\d+$/,
```
henrikstranneheim: qr/^\d+$/ -> qr/ ^\d+$ /xsm

t/peddy.t
```diff
@@ -0,0 +1,207 @@
...
+  MODULE:
+    for my $module (@modules) {
+        require_ok($module) or BAIL_OUT q{Cannot load} . $SPACE . $module;
+    }
+}
+
+use MIP::Program::Variantcalling::Peddy qw{ peddy };
+use MIP::Test::Commands qw{ test_function };
+
+diag(   q{Test peddy from Peddy v}
```
henrikstranneheim: Peddy.pm v

t/peddy.t
```diff
@@ -0,0 +1,207 @@
...
+    family_file_path => {
+        input => catfile(qw{ outfamily_file_directory family_id .fam }),
+        expected_output =>
+          catfile(qw{ outfamily_file_directory family_id .fam }),
+    },
+);
+
+my %specific_argument = (
+    processor_number => {
+        input           => 4,
```
henrikstranneheim: higher than 2 make it a CONSTANT

northwestwitch: Which syntax do you suggest in this case? Because if I write something like:
use constant N_PROCESSORS => 4;

And then I assign the parameter this way:
processor_number => {
        input           => N_PROCESSORS,
        expected_output => q{--procs 4},
    },

Then perlcritic complains, with the message: Pragma "constant" used at line 26, column 1.  See page 55 of PBP.  (Severity: 4) (It prints it in pink, even!)

northwestwitch: Well the expected output in this case should be:
expected_output => q{--procs} . N_PROCESSORS,

But still doesnät fix the perlcritic issue..

henrikstranneheim: Use:
use Readonly;
Readonly my $N_PROCESSORS => 4;

henrikstranneheim:
👍

PR #244, opened 2017-11-17
travis fix

jemten:


mip_install.pl
```diff
@@ -317,6 +318,14 @@
 
 $log->info( q{Writing install instructions to:} . $SPACE . $file_name_path );
 
+## Remove selected programs from installation
+if ( $parameter{skip_program} ) {
+    foreach my $skip ( @{ $parameter{skip_program} } ) {
```
henrikstranneheim: Add loop label

perl_install.pl
```diff
@@ -437,7 +444,7 @@ sub install_cpanm_modules {
     }
 
     ## Check aginst what's already installed
-    my @modules_to_install =
+    if my @modules_to_install =
```
henrikstranneheim: This looks odd? Why are you using an if here and without ()?

jemten: hmmm no reason to have it there whatsoever. Don't know what its doing there

henrikstranneheim:
Great

PR #253, opened 2017-11-23
Moved Bedtools intersectbed to MIP::Program::Variantcalling

northwestwitch:
Is it possible that intersectbed is just used by Bwa_rapid? How do I test it in rasta?


t/bedtools_intersectbed.t
```diff
@@ -0,0 +1,206 @@
...
+my %specific_argument = (
+    infile_path => {
+        input           => q{stdin},
+        expected_output => q{-a stdin},
+    },
+    outfile_path => {
+        input => catfile(
+            qw{ outsample_directory infile_prefix_sbatch_counteroutfile_suffix }
+        ),
+        expected_output => q{>}
```
henrikstranneheim: This should be removed right?

lib/MIP/Program/Variantcalling/Bedtools_intersectbed.pm
```diff
@@ -0,0 +1,142 @@
...
+        push @commands, q{-a} . $SPACE . $infile_path;
+    }
+
+    if ($intersectfile_path) {
+
+        push @commands, q{-b} . $SPACE . $intersectfile_path;
+    }
+
+    ## Outfile
+    if ($outfile_path) {
```
henrikstranneheim: You can remove this and use stdoutfile_path instead. Modify test and recipe accordingly

henrikstranneheim:
👍

PR #264, opened 2017-11-30
TIDDIT version update

jemten:


henrikstranneheim:
Add link to TIDDIT.py also

lib/MIP/Recipes/Install/Tiddit.pm
```diff
@@ -36,46 +36,54 @@ sub install_tiddit {
...
+
 
     my $tmpl = {
-        program_parameters_href => {
-            required    => 1,
-            default     => {},
+        conda_environment => {
             strict_type => 1,
-            store       => \$tiddit_parameters_href
+            store       => \$conda_environment
```
henrikstranneheim: Do not forget to add a "," for the last element or key i.e.: store       => \$conda_environment -> store       => \$conda_environment,

Perl will not mind not having a comma, unless you another element and then you will have subtle bug.

henrikstranneheim:
👍

PR #287, opened 2018-01-09
modified mip install to include vcf2cytosure installation

northwestwitch:
It works, the problem was the conda environment name, that on my computer is "base" instead of "root".

### DIFF
--- a/mip_install.pl
+++ b/mip_install.pl
@@ -50,6 +50,7 @@ use MIP::Recipes::Install::Sambamba qw{ install_sambamba };
 use MIP::Recipes::Install::SnpEff qw{ install_snpeff };
 use MIP::Recipes::Install::Svdb qw{ install_svdb };
 use MIP::Recipes::Install::Tiddit qw{ install_tiddit };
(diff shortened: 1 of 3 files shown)


mip_install.pl
```diff
@@ -386,6 +389,7 @@ sub build_usage {
     -cnvnr/--cnvnator_root_binary   Set the cnvnator root binary (Default: "root_v6.06.00.Linux-slc6-x86_64-gcc4.8.tar.gz")
     -tid/--tiddit                   Set the tiddit version (Default: "1.1.6")
     -svdb/--svdb                    Set the svdb version (Default: "1.0.6")
+    -v2cs/--vcf2cytosure            Set the svdb version (Default: "0.2.0")
```
dnil: svdb version -> vcf2cytosure. And looking good with MIP-vcf exported to cgi with current master; test in Cytosure expected tonight before new release.

northwestwitch: Fixed, thanks. And I guess when the new release is out the version will change as well!

henrikstranneheim:
👍

PR #291, opened 2018-01-11
Feature/mip c bwamem

hassanfa:


lib/MIP/Recipes/Pipeline/Mip_c.pm
```diff
@@ -180,88 +180,90 @@ sub pipeline_mip_c {
                 }
             );
         }
+
+# MultiQC
```
henrikstranneheim: Move this below markduplicates. Otherwise it will execute before there is data from bwa and markdup

henrikstranneheim:
👍

PR #293, opened 2018-01-11
Feature/mip c merge qc

hassanfa:
My last PR from before adding variant callers. Includes some basic QCs (without adapter trimming), alignment, merge, basic picard metric collector. Notably qccollect (yaml file was not available) and sambamba depth are missing.


henrikstranneheim:
👍

PR #303, opened 2018-01-16
Initial commit for Skewer, an adapter trimming tool.

hassanfa:


lib/MIP/Program/Trimming/Skewer.pm
```diff
@@ -0,0 +1,177 @@
...
+    my $infile_path;
+    my $outsuffix;
+    my $second_infile_path;
+    my $stderrfile_path;
+    my $stderrfile_path_append;
+    my $stdoutfile_path;
+    my $thread_number;
+    my $trim_mode;
+
+    ## Default(s)
```
henrikstranneheim: Remove

lib/MIP/Program/Trimming/Skewer.pm
```diff
@@ -0,0 +1,177 @@
...
+            strict_type => 1,
+        },
+        thread_number => {
+            allow       => qr/ ^\d+$ /xms,
+            default     => 2,
+            store       => \$thread_number,
+            strict_type => 1,
+        },
+        trim_mode => {
+            allow       => [qw{any ap head mp pe tail}],
```
henrikstranneheim: qw{any ap head mp pe tail} -> qw{ any ap head mp pe tail }
Always add the extra whitespace when using qw{} to be consisten within the MIP project

lib/MIP/Program/Trimming/Skewer.pm
```diff
@@ -0,0 +1,177 @@
...
+            default     => qw{pe},
```
henrikstranneheim: Even if only one element

lib/MIP/Program/Trimming/Skewer.pm
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    my @commands = q{skewer};
+
+    push @commands, q{-x} . $SPACE . $adapter_sequence;
+
+    # If a second read is provided, trim adapter accordingly
+    if ($second_infile_path) {
```
henrikstranneheim: if ($second_infile_path) { ->  if ($second_infile_path && $adapter_sequence_second) {

You seem to need both right? One to trigger and for the actual value

lib/MIP/Program/Trimming/Skewer.pm
```diff
@@ -0,0 +1,177 @@
...
+    my $tmpl = {
+        adapter_sequence => {
+            store       => \$adapter_sequence,
+            strict_type => 1,
+        },
+        adapter_sequence_second => {
+            store       => \$adapter_sequence_second,
+            strict_type => 1,
+        },
+        compress_output => {
```
henrikstranneheim: Is this a boolean?
allow and default would be nice

lib/MIP/Program/Trimming/Skewer.pm
hassanfa: There is a default value in Skewer for adapter sequences. I decided not to hardcode an adapter sequence into the module, but an explicit condition such as this can come handy later on.

lib/MIP/Program/Trimming/Skewer.pm
hassanfa: This one is also updated in the new commit, but github doesn't show it as changed in my PR.

henrikstranneheim:
👍

PR #307, opened 2018-01-17
Test for Record add_gene_panel

northwestwitch:
The other tests are already done

### DIFF
--- a/t/record_add_gene_panel.t
+++ b/t/record_add_gene_panel.t
@@ -118,6 +118,24 @@ add_gene_panel(
 is( exists $sample_info{$program_name_test}{$aggregate_gene_panels_key}{gene_panel},
     1, q{Gene panel added to $sample_info} );
 
+is( $sample_info{$program_name_test}{$aggregate_gene_panels_key}{gene_panel}{q{gene_panel}}, q{gene_panel_test}, q{Gene panel added to $sample_info} );
+
+#print "\n----->".$sample_info{$program_name_test}{$aggregate_gene_panels_key}{gene_panel};
+
+
+while(my($key1,$value1)=each(%sample_info)){
+  while(my($key2,$value2)=each(%$value1)){
+    print "first level,key1:$key1\n";
+    print "\tsecond level, key2:$key2,value2-->$value2\n";
+    if ($key2 eq "TEST"){
+      foreach my $key3 (keys %$value2)
+      {
+        print "\t\t\tkey3 is $key3\n";
+      }
+    }
+  }
+}
+
 
 done_testing();
 
--- a/t/record_add_gene_panel.t
+++ b/t/record_add_gene_panel.t
@@ -99,6 +99,7 @@ diag(   q{Test add_gene_panel from Record.pm v}
       . $SPACE
       . $EXECUTABLE_NAME );
 
+my $aggregate_gene_panel_file = catfile( $Bin, qw{ data 643594-miptest aggregated_gene_panel_test.txt } );
 my $aggregate_gene_panels_key = q{TEST};
 my $family_id_test = q{family_id};
 my $program_name_test = q{vcfparser};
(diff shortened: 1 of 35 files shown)


t/record_add_program_outfile_to_sample_info.t
```diff
@@ -120,10 +120,54 @@ is( $sample_info{program}{$test_program_name}{path},
...
+);
+## Test
+is(
+    !exists $sample_info{sample}{$sample_id}{program}{$test_program_name}
+      {$infile},
+    1,
+    q{Sample level hash key not created when infile is not provided}
+);
+
+is(
```
henrikstranneheim: You should be able to loop over this. Something like:
my %test_no_infile = (outdirectory => q{Value to sample level outdirectory not assigned});
'#' and so on for the parameters you want to test and then:

while (my ($parameter, $test_comment) = each %test_no_infile) {
my $test_result = $sample_info{sample}{$sample_id}{program}{$test_program_name}
+      {$infile}{$parameter};
is($test_result, undef, $test_comment);
}

northwestwitch:
I've realized that while I can loop over the un-assigned params when you are not using an infile (because they all return undef), you can't do the same for when you have the infile (because you expect defined values, one for each cycle). So I left the last part of the code as it was.

t/record_add_program_outfile_to_sample_info.t
```diff
@@ -120,10 +120,40 @@ is( $sample_info{program}{$test_program_name}{path},
...
+    }
+);
+## Test
+
+is( $sample_info{sample}{$sample_id}{program}{$test_program_name}{path},
+    $path, q{Assigned correct value to sample level path} );
+
+my %test_no_infile =
+  ( outdirectory => q{Value to sample level outdirectory not assigned} );
+$test_no_infile{outfile} .= q{Value to sample level outfile not assigned};
```
henrikstranneheim: Continue with the key value assignment from the initial my instead. Like this:
my %test_no_infile = ( outdirectory => q{Value to sample level outdirectory not assigned},
                                     outfile => q{Value to sample level outfile not assigned},
                                     path => q{Value to sample level path not assigned},
                                      version => q{Value to sample level version not assigned},
                                   );

t/record_add_gene_panel.t
```diff
@@ -0,0 +1,155 @@
...
+    {
+      aggregate_gene_panel_file => $aggregate_gene_panel_file,
+      aggregate_gene_panels_key => $aggregate_gene_panels_key,
+      family_id                 => $family_id_test,
+      program_name => $program_name_test,
+      sample_info_href => \%sample_info,
+    }
+);
+
+is( exists $sample_info{$program_name_test}{$aggregate_gene_panels_key}{gene_panel},
```
henrikstranneheim: You are only testing the existence of the key here and not the value

also add a test for the value:
is( exists $sample_info{$program_name_test}{$aggregate_gene_panels_key}{gene_panel}, q{gene_panel_test}, q{Gene panel added to $sample_info} );

Double check that the expected return actually is q{gene_panel_test}.

northwestwitch: It's actually not q{gene_panel_test}, it's a hash.

If I print this hash of hashes I get this:
first level,key1:TEST
	second level, key2:path,value2-->path_to_gene_panel.txt (shortened by me)
first level,key1:vcfparser
	second level, key2:TEST,value2-->HASH(0x7fbffacae400)

I'm not really understanding what I'm supposed to retrieve here..

henrikstranneheim:
@northwestwitch Can you pull these changes to your local branch and see if they make pass.

northwestwitch:
Yes, now I see. Didn't think of using the header info to provide the third level keys. I'll work on the other test now!

henrikstranneheim:
Excellent!

PR #320, opened 2018-01-18
test for quote_bash_variable

jemten:


henrikstranneheim:
I was kind of suprised that you used the test_command.t template and not the test.t template as this is not really a command even though the actual test looks fine. Perhaps for consistency you should modify it as a standard sub test instead of using the test command structure. e.g
my return_quoted = sub(string_with_variable_to_quote => q{$test});
is($return_quoted, q{"$test"}, q{Comment]);

henrikstranneheim:
👍

PR #322, opened 2018-01-18
Feature/test path bed file exists

northwestwitch:
I chose this test because it looked like it was an easy one, but I've spent the whole afternoon on google to understand why it doesn't work.
My idea was to use Perl::Trap to trap the exit signal sent when the file to be tested has the wrong format, but I had no luck. Now I'm curious to see how you would solve the issue :)

### DIFF
--- a/lib/MIP/Check/Path.pm
+++ b/lib/MIP/Check/Path.pm
@@ -23,7 +23,7 @@ BEGIN {
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK =
-      qw{ check_filesystem_objects_existance check_filesystem_objects_and_index_existance check_file_version_exist check_dir_path_exist check_target_bed_file_exist check_parameter_files };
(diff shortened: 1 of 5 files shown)


henrikstranneheim:
The trick is that you do not have to test for fail. This is done by the sub itself at runtime. You only have to test that is allows a correct file to pass through and does not fail everything.

northwestwitch:
Ok, I removed the test fail check, but still would have been nice to Trap the exit code. Bah!

henrikstranneheim:
You do not even have to create a temp file, a simple string in to the sub will suffice.

northwestwitch:
Yes, but this way it looked more professional! Do you want me to change it?

henrikstranneheim:
Yes, more professional indeed 😄 !
However,  I think the sub is named improperly it should be: "check_target_file_bed_suffix".
The name check_target_bed_file_exist and give the wrong impression of what the sub really does i.e. that is handles files and not just checks strings.

Could you please rename it and remove the temp file and use a string instead. That would be great.

henrikstranneheim:
Beautiful! 👍

PR #336, opened 2018-01-22
Feature/test check references for vt

northwestwitch:
Tested for all types of program parameters (SCALAR, ARRAY and HASH). Is this test ok or should I test something else as well?

### DIFF
--- a/t/check_references_for_vt.t
+++ b/t/check_references_for_vt.t
@@ -116,35 +116,28 @@ my $log = initiate_logger(
 );
 
 my %active_parameter_test = (
-    gatk_baserecalibration_known_sites => [
-        qw{ GRCh37_dbsnp_-138-.vcf GRCh37_1000g_indels_-phase1-.vcf GRCh37_mills_and_1000g_indels_-gold_standard-.vcf },
-    ],
-    gatk_realigner_indel_known_sites => [
-        qw{ GRCh37_1000g_indels_-phase1-.vcf GRCh37_mills_and_1000g_indels_-gold_standard-.vcf },
-    ],
-    gatk_varianteval_dbsnp   => q{GRCh37_dbsnp_-138_esa_129-.vcf},
     pgatk_baserecalibration          => 1,
+    gatk_baserecalibration_known_sites => [catfile($Bin, qw{ data references GRCh37_dbsnp_-138-.vcf }) , catfile($Bin, qw{ data references GRCh37_1000g_indels_-phase1-.vcf }), catfile($Bin, qw{ data references GRCh37_mills_and_1000g_indels_-gold_standard-.vcf }) ],
     pgatk_realigner                  => 1,
+    gatk_realigner_indel_known_sites => [ catfile($Bin, qw{ data references GRCh37_1000g_indels_-phase1-.vcf }), catfile($Bin, qw{ data references GRCh37_mills_and_1000g_indels_-gold_standard-.vcf })],
     pgatk_variantevalall             => 1,
+    gatk_varianteval_dbsnp           => catfile($Bin, qw{ data references GRCh37_dbsnp_-138_esa_129-.vcf }),
     pgatk_variantevalexome           => 1,
     psnpeff                          => 1,
     snpsift_annotation_files => {
-        SWEREF => catfile($Bin, qw{ t data references GRCh37_anon-swegen_snp_-1000samples-.vcf.gz }),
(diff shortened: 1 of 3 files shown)


t/check_references_for_vt.t
```diff
@@ -0,0 +1,200 @@
...
+);
+
+my %active_parameter_test = (
+    pgatk_realigner                  => 1,
+    pgatk_baserecalibration          => 1,
+    pgatk_variantevalall             => 1,
+    pgatk_variantevalexome           => 1,
+    psnpeff                          => 1,
+    gatk_realigner_indel_known_sites => [
+        qw{ GRCh37_1000g_indels_-phase1-.vcf GRCh37_mills_and_1000g_indels_-gold_standard-.vcf },
```
henrikstranneheim: You need to add the mip_reference file path to the annotation file. This is done before entering into this check. You can use: catfile($Bin, qw{ t data references }).

Otherwise you are just testing this:
sub check_if_processed_by_vt 
if ( not -e $reference_file_path ) {

        ## Do nothing since there is not ref file to check
        return;
    }

t/check_references_for_vt.t
```diff
@@ -0,0 +1,200 @@
...
+    ],
+    gatk_baserecalibration_known_sites => [
+        qw{ GRCh37_dbsnp_-138-.vcf GRCh37_1000g_indels_-phase1-.vcf GRCh37_mills_and_1000g_indels_-gold_standard-.vcf },
+    ],
+    gatk_varianteval_dbsnp   => q{GRCh37_dbsnp_-138_esa_129-.vcf},
+    snpsift_annotation_files => {
```
henrikstranneheim: This file is good to test as it actually has the vcf keys we are looking for: t/data/references/GRCh37_gnomad.genomes_-r2.0.1-.vcf.gz

t/check_references_for_vt.t
```diff
@@ -0,0 +1,200 @@
...
+    }
+);
+
+my %active_parameter_test = (
+    pgatk_realigner                  => 1,
+    pgatk_baserecalibration          => 1,
+    pgatk_variantevalall             => 1,
+    pgatk_variantevalexome           => 1,
+    psnpeff                          => 1,
+    gatk_realigner_indel_known_sites => [
```
henrikstranneheim: Order this AB

t/check_references_for_vt.t
```diff
@@ -0,0 +1,200 @@
...
+        associated_program =>
+          [qw{ pgatk_variantevalall pgatk_variantevalexome }],
+        data_type => q{SCALAR},
+    },
+    snpsift_annotation_files =>
+      { associated_program => [qw{ psnpeff }], data_type => q{HASH}, },
+);
+
+my @vt_references_test =
+  qw{ gatk_realigner_indel_known_sites gatk_baserecalibration_known_sites gatk_varianteval_dbsnp snpsift_annotation_files };
```
henrikstranneheim: Order AB

t/check_references_for_vt.t
```diff
@@ -0,0 +1,200 @@
...
+my @refs_to_process = check_references_for_vt(
+    {
+        active_parameter_href => \%active_parameter_test,
+        log                   => Log::Log4perl->get_logger(q{TEST}),
+        parameter_href        => \%parameter_test,
+        vt_references_ref     => \@vt_references_test,
+    }
+);
+
+is( @refs_to_process, 0, q{Test passed, 0 references to process.} );
```
henrikstranneheim: This is probably since you do not check the files yet as there is no ref path supplied

t/check_references_for_vt.t
northwestwitch: So, should I keep this test only and remove the others, where I have no annotation file under the t/data/references/ folder?

henrikstranneheim: No keep them all, but add gnomad

northwestwitch:
Ok.

t/check_references_for_vt.t
northwestwitch: I tried to fix it according to your comments but I don't think I'm doing it right. I still get 0 as number of references to process. The thing is I don't really understand the code I'm trying to test. What is that it does, what do you mean by "references to process", there reference files are supposed to be found in that folder or not?

henrikstranneheim:
I think it looks good. I will merge this and we can go through it tomorrow.

PR #355, opened 2018-01-25
Changed VEP plugin installation dir to cache dir, fixes #353

### LINKED ISSUE #353: VEP: Add "Plugins" to cache dir path when adding data for plugins
In install recipe:
ensembl-tools-release-91/cache -> ensembl-tools-release-91/cache/Plugins

when adding data to plugins.

jemten:


henrikstranneheim:
👍

Issue #371, opened 2018-01-26
Vardict post process filtering


hassanfa:
Series of bcftools filter strings and custom sed/awk filter to clean out VCF file and tag variants. http://bcb.io/2016/04/04/vardict-filtering/ tied to #357

Issue #357, opened 2018-01-25
extra filters for var2vcf_paired


hassanfa:
Vardict recipe's output is an unsorted and unfiltered VCF file. Following filters (or similar) need to be applied:

- [x] var2vcf_paired p-value: To increase number of variants kept for post processing
- [x] var2vcf_paired maximum mismatches allowed: To control false positives
- [x] var2vcf_paired select only somatic variants: obviously!
- [x] Update vardict recipe
- [x] Update define_parameters yaml file
- [x] Update mip.pl usage
- [x] Update mip.pl command line options

PR #372, opened 2018-01-26
Feature/vardict_postprocessing

hassanfa:
PR for Vardict's post processing to address issues #371 and #357 , and using filter_mode update to bcftools filter PR #370


lib/MIP/Recipes/Analysis/Vardict.pm
```diff
@@ -408,4 +475,131 @@ sub analysis_vardict_tn {
...
+    ## Function: Create filter expression for bcftools.
+    ## Return: $bcftools_filter_expr
+
+    ## The filter is: AF < 0.2 && QUAL < 55 && SSF > 0.06
+
+    Readonly my $CONDITON_AND => q{&&};
+
+    my $bcftools_filter_expr;
+
+    my $af_expr = q{AF < 0.2};
```
henrikstranneheim: I think it would be even better if you set the values as constants as well since they are hard-coded i.e.
Readonly my $ALLELE_FREQ => 0.2
etc

lib/MIP/Recipes/Analysis/Vardict.pm
```diff
@@ -408,4 +475,131 @@ sub analysis_vardict_tn {
...
+      . $dp_expr
+      . $CONDITON_OR
+      . $qual_expr
+      . $CLOSE_PARAN
+      . $CLOSE_PARAN;
+
+    return $bcftools_filter_expr;
+}
+
+sub _filter_duplicate {
```
henrikstranneheim: Could you use vt uniq or bcftools norm --remove-duplicates here instead of a custom awk?

lib/MIP/Recipes/Analysis/Vardict.pm
```diff
@@ -408,4 +475,131 @@ sub analysis_vardict_tn {
...
+}
+
+sub _replace_ambiguous {
+
+    ## Function: Writes a awk expression to open $FILEHANDLE. The awk expression will replace ambiguous reference with N. Adopted from bcbio.
+    ## Returns: $awk_filter
+
+    ## Remove abimguous from ref, column 4 in VCF
+    my $awk_filter_ref =
+        q?awk -F$'\t' -v OFS='\t' ' ?
```
henrikstranneheim: Please comments this more. Preferebly for each argument

lib/MIP/Recipes/Analysis/Vardict.pm
hassanfa: I'll include rational for each hard-coded value in my next commit. Since these values are quite diverse, I had them match the references provided. If this doesn't satisfy common MIP structure, I'll swap them to constants.

lib/MIP/Recipes/Analysis/Vardict.pm
hassanfa: Aren't those two filtering two recurring lines rather than columns? This is for specifically checking if alt and ref allele are same, and if they are same it will be removed.

lib/MIP/Recipes/Analysis/Vardict.pm
hassanfa: Will do.

lib/MIP/Recipes/Analysis/Vardict.pm
henrikstranneheim: Ok

lib/MIP/Recipes/Analysis/Vardict.pm
henrikstranneheim: OK, good.

lib/MIP/Recipes/Analysis/Vardict.pm
```diff
@@ -578,26 +566,26 @@ sub _filter_duplicate {
...
-    ##        - "{ print }": Print!
-
     ## Remove abimguous from ref, column 4 in VCF
-    my $awk_filter_ref =
-        q?awk -F$'\t' -v OFS='\t' ' ?
-      . q?{ if ($0 !~ /^#/) gsub(/[KMRYSWBVHDXkmryswbvhdx]/, "N", $4)} ?
-      . q?{ print }'?;
+    my $awk_filter_ref = q?awk -F$'\t' -v OFS='\t' ' ?
+
+      #Lines not starting with "#" essentially the VCF header.
```
henrikstranneheim: Always add a whitespace between '#' and the text

henrikstranneheim:
👍

PR #387, opened 2018-02-02
changed the default infile command

J35P312:
Hello!
Nicely done, the star_aln recipe works great!

I did however confuse gzip for gunzip as an infiles command. This should do the trick!

Have a nice weekend!
//JEsper


henrikstranneheim:
Great. Let's discuss on monday

PR #394, opened 2018-02-06
Now linking to environment specific library

jemten:
My git clone has stopped working again so I haven't been able to test it. Should work though

### DIFF
--- a/lib/MIP/Recipes/Install/Vep.pm
+++ b/lib/MIP/Recipes/Install/Vep.pm
@@ -21,7 +21,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.07;
+    our $VERSION = 1.08;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{ install_vep };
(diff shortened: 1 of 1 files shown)


lib/MIP/Recipes/Install/Vep.pm
```diff
@@ -215,13 +214,8 @@ q{Please add the [a] and/or [l] flag to --vep_auto_flag when running mip_install
     }
 
     ## Set LD_LIBRARY_PATH for VEP isntallation
```
henrikstranneheim: isntallation

lib/MIP/Recipes/Install/Vep.pm
```diff
@@ -215,13 +214,8 @@ q{Please add the [a] and/or [l] flag to --vep_auto_flag when running mip_install
...
-    my $conda_dir_path = get_conda_dir_path(
-        {
-            log => $log
-        }
-    );
     say $FILEHANDLE q{LD_LIBRARY_PATH=}
-      . $conda_dir_path
+      . $conda_prefix_path
```
henrikstranneheim: How does this relate to the $conda_environment that is part of the args for the installation recipe? Will this be dynamic so that no conda env will point to base and otherwise to env?

jemten: $conda_prefix_path is the full path to the conda environment that is created during the installation. If no env is given on the command line the variable will point to the base environment. The variable $conda_environment contains the name of the environment created and is used when activating the env during the installation.

henrikstranneheim:
👍

PR #395, opened 2018-02-06
Feature/general_get

hassanfa:
Added a general module to retrieve information from pedigree hash ref. It is needed to retrieve samples in certain order for cancer pipeline.


t/get_pedigree_sample_info.t
```diff
@@ -0,0 +1,275 @@
...
+
+    ## Function :   Check if all elements of @output_info are in @sample_info
+    ## Returns  :   $success
+
+    my $success = 1;
+    if ( scalar @sample_info eq scalar @output_info ) {
+
+        foreach my $expected (@output_info) {
+
+            if ( not grep $expected eq $_, @sample_info ) {
```
hassanfa: perlcritic reports this as severity 4:

```
"grep" used in boolean context at line 262, column 22.  See pages 71,72 of PBP.  (Severity: 2)
Expression form of "grep" at line 262, column 22.  See page 169 of PBP.  (Severity: 4)
```

henrikstranneheim: Instead, a better solution is to use the any function from List::MoreUtils, which short-circuits after the first successful match to save time.

henrikstranneheim: from perl critic

lib/MIP/Get/Pedigree.pm
```diff
@@ -0,0 +1,99 @@
...
+    # Functions and variables which can be optionally exported
+    our @EXPORT_OK = qw{ get_sample_info };
+}
+
+sub get_sample_info {
+
+    ## Function     : Given pedigree hash, extract an specific field given a sample level pedigree info
+    ## Returns      : @sample_info
+    ## Arguments    : $pedigree_hash        =>  Pedigree hash {REF}
+    ##              : $sample_info_key      =>  Key in pedigree hash to search
```
henrikstranneheim: These are used to limit the output to if certain conditions is meet. It would be nice if this was evident from the name e.g. sample_info_intersect_key and sample_info_intersect_value

lib/MIP/Get/Pedigree.pm
```diff
@@ -0,0 +1,99 @@
...
+    our $VERSION = 1.01;
+
+    # Functions and variables which can be optionally exported
+    our @EXPORT_OK = qw{ get_sample_info };
+}
+
+sub get_sample_info {
+
+    ## Function     : Given pedigree hash, extract an specific field given a sample level pedigree info
+    ## Returns      : @sample_info
```
henrikstranneheim: @info_out

lib/MIP/Get/Pedigree.pm
```diff
@@ -0,0 +1,99 @@
...
+    ## Arguments    : $pedigree_hash        =>  Pedigree hash {REF}
+    ##              : $sample_info_key      =>  Key in pedigree hash to search
+    ##              : $sample_info_value    =>  Key in pedigree hash to match
+    ##              : $sample_info_out      =>  Key in pedigree hash to return
```
henrikstranneheim: $sample_info_out -> get_values_for_key

lib/MIP/Get/Pedigree.pm
```diff
@@ -0,0 +1,99 @@
...
+        pedigree_href => {
+            default     => {},
+            defined     => 1,
+            required    => 1,
+            store       => \$pedigree_href,
+            strict_type => 1,
+        },
+        sample_info_key => {
+            defined     => 1,
+            required    => 1,
```
henrikstranneheim: losen this so that they are not required. That way you can restrict the values returned to s single sample if you wish or all samples.

lib/MIP/Get/Pedigree.pm
```diff
@@ -0,0 +1,99 @@
...
+        },
+        sample_info_out => {
+            defined     => 1,
+            required    => 1,
+            store       => \$sample_info_out,
+            strict_type => 1,
+        },
+        sample_info_value => {
+            defined     => 1,
+            required    => 1,
```
henrikstranneheim: losen this so that they are not required. That way you can restrict the values returned to s single sample if you wish or all samples,

t/get_pedigree_sample_info.t
```diff
@@ -0,0 +1,276 @@
...
+
+    return <<"END_USAGE";
+ $program_name [options]
+    -vb/--verbose Verbose
+    -h/--help Display this help message
+    -v/--version Display version
+END_USAGE
+}
+
+sub _compare_aray {
```
henrikstranneheim: array

t/get_pedigree_sample_info.t
```diff
@@ -0,0 +1,276 @@
...
+        pedigree_href     => \%pedigree,
+        sample_info_key   => q{sample_origin},
+        sample_info_out   => q{sample_id},
+        sample_info_value => q{tumor},
+    }
+);
+
+my @output_info = (qw{sample_2_a});
+
+is( _compare_aray(), 1, "Test for a single info retrieval." );
```
henrikstranneheim: use http://search.cpan.org/~zmij/Array-Utils-0.5/Utils.pm instead

t/get_pedigree_sample_info.t
```diff
@@ -0,0 +1,276 @@
...
+diag(   q{Test get_sample_info from Pedigree.pm v}
+      . $MIP::Get::Pedigree::VERSION
+      . $COMMA
+      . $SPACE . q{Perl}
+      . $SPACE
+      . $PERL_VERSION
+      . $SPACE
+      . $EXECUTABLE_NAME );
+
+## Create temp logger
```
henrikstranneheim: Do you need the log?

lib/MIP/Get/Pedigree.pm
hassanfa: I'm thinking for a good way to implement what you mentioned here, but I can't properly sketch it in my head. Since I'm looking for (key, value) pairs in samples level of pedigree_href, the value has to be provided by user. So you're suggesting I have to e.g. return an array of sample_ids if no pair of (key, value) is not provided?

henrikstranneheim:
Closing this. Reopen if required in the future

PR #405, opened 2018-02-12
Added vep and svdb to conflicting programs array

jemten:


henrikstranneheim:
Could you also include an update to the develop docs README

PR #410, opened 2018-02-27
fixed MT depth calculation and updated default MT subsampling depth, …

### LINKED ISSUE #409: Fix Samtools_subsample_MT.pm
The coverage of the subsampled MT bam file does not match the expected value

jemten:
…fixes #409

### DIFF
--- a/lib/MIP/Recipes/Analysis/Samtools_subsample_MT.pm
+++ b/lib/MIP/Recipes/Analysis/Samtools_subsample_MT.pm
@@ -248,8 +248,9 @@ sub analysis_samtools_subsample_MT {
     # Get depth per base
     samtools_depth(
         {
-            FILEHANDLE  => $FILEHANDLE,
-            infile_path => $infile_path,
+            FILEHANDLE         => $FILEHANDLE,
+            infile_path        => $infile_path,
+            max_depth_treshold => q{500000},
         }
     );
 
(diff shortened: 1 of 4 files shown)


t/samtools_depth.t
```diff
@@ -142,6 +142,10 @@ my %specific_argument = (
         input           => q{infile.test},
         expected_output => q{infile.test},
     },
+    max_depth_treshold => {
+        input           => q{100000},
```
henrikstranneheim: Should this not be a number and made into a constant with Readonly and use of "_" a separator for large numbers?

lib/MIP/Recipes/Analysis/Samtools_subsample_MT.pm
```diff
@@ -248,8 +248,9 @@ sub analysis_samtools_subsample_MT {
     # Get depth per base
     samtools_depth(
         {
-            FILEHANDLE  => $FILEHANDLE,
-            infile_path => $infile_path,
+            FILEHANDLE         => $FILEHANDLE,
+            infile_path        => $infile_path,
+            max_depth_treshold => q{500000},
```
henrikstranneheim: Should this not be a number and made into a constant with Readonly and use of "_" a separator for large numbers?

jemten:
Will reject this pull request and submit a new one to the correct branch

PR #414, opened 2018-02-28
Star fusion analysis recipe

J35P312:
Hello there!
Here's the first commit of the star fusion analysis recipie


henrikstranneheim:
Will fix in feature branch instead

Issue #429, opened 2018-03-16
Failing tests


jemten:
- [x] check_if_processed_by_vt.t
  - Requires a somewhat recent version of bcftools to work
- [x] contigs_set_contigs.t
- [x] get_capture_kit.t
- [x] mip.t 
- [x] mip_install.t
- [x] set_custom_default_to_active_parameter.t
  - solved in commit 596f044c390163d77f7ae6ac442c15e709399713
- [x] update_to_absolute_path.t
- [x] set_pedigree_capture_kit_info.t

PR #431, opened 2018-03-19
fix for the set contig test, #429

jemten:
No more special case when setting contigs for structural variants


henrikstranneheim:
👍

PR #438, opened 2018-03-21
fixed broken set_pedigree_capture_kit_info.t, #429

jemten:


jemten:
Hmm, might have misunderstood the function a bit. Does it need to return undef for this case?

t/set_pedigree_capture_kit_info.t
```diff
@@ -234,8 +234,12 @@ set_pedigree_capture_kit_info(
         user_supply_switch_href => \%user_supply_switch,
     }
 );
-is( $active_parameter{exome_target_bed},
-    undef, q(No capture kit from cmd, config or pedigree) );
+
+is(
```
henrikstranneheim: It should return undef to signify that the user has used a capture kit where there is no akronym shortcut that can be used. Actually setting to latest is done elsewhere and occurs when the user did not set any capture kit info at all in this run or previous runs. It is a bit confusing I know.

lib/MIP/Set/Pedigree.pm
```diff
@@ -195,11 +195,11 @@ sub set_pedigree_capture_kit_info {
         }
 
         ## No recorded capture kit from pedigree or previous run
-	if ( not $capture_kit ) {
+	    if ( not $capture_kit ) {
```
henrikstranneheim: Is the indent really correct?

jemten: It isn't. Though it looks coorect on my local copy...

t/set_pedigree_capture_kit_info.t
jemten: OK. Right now it looks like it is set to latest in the set_pedigree_capture_kit_info subroutine

henrikstranneheim: Yes, you are rigth I confused it with get_capture_kit. When there is no info from cmd, config or pedigree we should default to "last" to set the default.

lib/MIP/Set/Pedigree.pm
henrikstranneheim: ok

henrikstranneheim:
👍

PR #461, opened 2018-04-11
added codna-forge, fix #366 and fix #448

### LINKED ISSUE #366: Proper setup of conda channels

### LINKED ISSUE #448: Update the conda install sub
Add conda-forge as a channel (after bioconda) when downloading bioconda packages

jemten:
Added conda-forge as a channel when installing bioconda packages

### DIFF
--- a/lib/MIP/Package_manager/Conda.pm
+++ b/lib/MIP/Package_manager/Conda.pm
@@ -1,18 +1,18 @@
 package MIP::Package_manager::Conda;
 
 use Carp;
-use charnames qw{ :full :short };
 use Cwd;
 use English qw{ -no_match_vars };
 use Getopt::Long;
 use IO::Handle;
+use List::Util qw{ any };
 use Params::Check qw{ check allow last_error };
-$Params::Check::PRESERVE_CASE = 1;    #Do not convert to lower case
+use charnames qw{ :full :short };
 use open qw{ :encoding(UTF-8) :std };
 use strict;
 use utf8;
-use warnings;
 use warnings qw{ FATAL utf8 };
+use warnings;
 
 ## Cpanm
 use IPC::Cmd qw{ can_run run };
@@ -31,7 +31,7 @@ BEGIN {
     require Exporter;
 
     # Set the version for version checking
-    our $VERSION = 1.09;
+    our $VERSION = 1.10;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK =
@@ -42,17 +42,17 @@ sub conda_create {
 
 ##Function : Create Conda environment
 ##Returns  : @commands
-##Arguments: $conda_channel   => Search for packages in specified conda channel
-##         : $env_name        => Name of environment to create
-##         : $FILEHANDLE      => Filehandle to write to
-##         : $no_confirmation => Do not ask for confirmation
-##         : $packages_ref    => Packages to be installed
-##         : $quiet           => Do not display progress bar
+##Arguments: $conda_channels_ref => Search for packages in specified conda channels {REF}
+##         : $env_name           => Name of environment to create
+##         : $FILEHANDLE         => Filehandle to write to
+##         : $no_confirmation    => Do not ask for confirmation
+##         : $packages_ref       => Packages to be installed
+##         : $quiet              => Do not display progress bar
 
     my ($arg_href) = @_;
 
     ## Flatten argument(s)
-    my $conda_channel;
+    my $conda_channels_ref;
     my $env_name;
     my $FILEHANDLE;
     my $no_confirmation;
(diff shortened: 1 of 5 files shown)


lib/MIP/Package_manager/Conda.pm
```diff
@@ -1,18 +1,18 @@
...
 use Carp;
-use charnames qw{ :full :short };
 use Cwd;
 use English qw{ -no_match_vars };
 use Getopt::Long;
 use IO::Handle;
+use List::Util qw{ any };
 use Params::Check qw{ check allow last_error };
-$Params::Check::PRESERVE_CASE = 1;    #Do not convert to lower case
+use charnames qw{ :full :short };
```
henrikstranneheim: Why not in AB?

lib/MIP/Package_manager/Conda.pm
```diff
@@ -42,54 +42,66 @@ sub conda_create {
...
     my $no_confirmation;
     my $packages_ref;
     my $quiet;
 
     my $tmpl = {
-        conda_channel => {
-            defined     => 1,
+        conda_channels_ref => {
+            allow => sub {
+                my $channels_ref = shift;
```
henrikstranneheim: Maybe you should be more explicit and write out what youa re "shifting" i.e. shift @_

lib/MIP/Package_manager/Conda.pm
```diff
@@ -42,54 +42,66 @@ sub conda_create {
...
+                ## Allow undef
+                return 1 if scalar @{$channels_ref} == 0;
+                ## Test values
+                foreach my $channel ( @{$channels_ref} ) {
```
henrikstranneheim: Consider making this loop a utility sub, so that it can be labelel with a name for its function and tested (if required). Also add a loop label

lib/MIP/Package_manager/Conda.pm
```diff
@@ -325,56 +340,68 @@ sub conda_install {
...
+##         : $FILEHANDLE         => Filehandle to write to
+##         : $no_confirmation    => Do not ask for confirmation
+##         : $packages_ref       => Packages to be installed
+##         : $quiet              => Do not display progress bar
 
     my ($arg_href) = @_;
 
     ## Flatten argument(s)
-    my $conda_channel;
+    my $conda_channels_ref = $arg_href->{conda_channels_ref};
```
henrikstranneheim: Move this under ## default as my $conda_channels_ref;
Could this not be set in the $tmpl? with default => $arg_href->{conda_channels_ref}

lib/MIP/Package_manager/Conda.pm
```diff
@@ -325,56 +340,68 @@ sub conda_install {
...
-        conda_channel => {
-            defined     => 1,
+        conda_channels_ref => {
+            default => [],
+            allow   => sub {
+                my $channels_ref = shift;
+                ## Allow undef
+                return 1 if scalar @{$channels_ref} == 0;
+                ## Test values
+                foreach my $channel ( @{$channels_ref} ) {
```
henrikstranneheim: Since you always seem to want to add conda-forge bioconda. Could it be simpler to just:
uniqe push @{$channels_ref}, [qw{ conda-forge bioconda }];

Also if you want bioconda to take priority over conda-forge just add them in the reverse order, i.e.  [qw{ bioconda conda-forge }]. Then you do not have to sort later. It could be confusing if one supplies one order and gets back a changed one if the order of the element is the priority. Maybe the user of the sub wants forge before bioconda (for some obscure reason). Maybe I am missing some subtlety here.

lib/MIP/Package_manager/Conda.pm
```diff
@@ -396,17 +423,20 @@ sub conda_install {
         push @commands, q{--yes};
     }
 
-    if ($conda_channel) {
-        push @commands, q{--channel} . $SPACE . $conda_channel;
+    if ( @{$conda_channels_ref} ) {
+        ## Push channels in reverse order so that bioconda gets priority over conda-forge
+        push @commands,
+          q{--channel} . $SPACE . join $SPACE . q{--channel} . $SPACE,
+          sort @{$conda_channels_ref};
```
henrikstranneheim: See comment above

lib/MIP/Package_manager/Conda.pm
jemten: Maybe. Bioconda needs to have priority over conda-forge. Either we always add conda-forge to bioconda in the command routine like this 
```Perl
if ( @{$conda_channels_ref} ) {
  ## Add bioconda and conda-forge so that bioconda gets priority over conda-forge
  if ( grep { $_ eq q{bioconda} } @{ $conda_channels_ref } ) {
    @{ $conda_channels_ref } = uniq (qw{ bioconda conda-forge}, @{ $conda_channels_ref });
  }
  push @commands, q{--channel} . $SPACE . join $SPACE . q{--channel} . $SPACE, @{$conda_channels_ref};
}
```
or we leave it to the one coding to supply the channels in the correct order when calling the subroutine.  Maybe that is the best option.

jemten: Then the subroutine checking for allowed conda channels might be overkill as well.

lib/MIP/Package_manager/Conda.pm
```diff
@@ -533,4 +531,43 @@ sub conda_uninstall {
     return @commands;
 }
 
+sub _check_array_membership {
+
+##Function : Checks if all array elements are part of an array with allowed elements. Returns tru/false
```
henrikstranneheim: ##F -> ## F
tru -> true

lib/MIP/Package_manager/Conda.pm
```diff
@@ -107,6 +106,8 @@ sub conda_create {
 
     check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
 
+    use List::Util qw{ uniq };
```
henrikstranneheim: Are you using this import?

henrikstranneheim:
👍

PR #462, opened 2018-04-12
no user prompt when not installing latest release, fixes #457

### LINKED ISSUE #457: Update VEP install options
Add VEP version and VEP cache version to the VEP install command

jemten:


lib/MIP/Program/Variantcalling/Vep.pm
```diff
@@ -256,37 +254,47 @@ sub variant_effect_predictor {
...
+    my $version;
 
     my $tmpl = {
         assembly        => { store => \$assembly,        strict_type => 1, },
         auto            => { store => \$auto,            strict_type => 1, },
         cache_directory => { store => \$cache_directory, strict_type => 1, },
+        cache_version   => { store => \$cache_version,   strict_type => 1, },
         FILEHANDLE      => { store => \$FILEHANDLE, },
+        no_update       => {
+            allow       => [ undef, 0, 1 ],
```
henrikstranneheim: Any default setting here? You seem to have locked it at 1 by hardcoding it in the sub call and not using a user option

lib/MIP/Program/Variantcalling/Vep.pm
```diff
@@ -298,6 +306,8 @@ sub variant_effect_predictor_install {
         stderrfile_path_append =>
           { store => \$stderrfile_path_append, strict_type => 1, },
         stdoutfile_path => { store => \$stdoutfile_path, strict_type => 1, },
+        version         => { store => \$version,         strict_type => 1, },
+
```
henrikstranneheim: Remove newline

t/variant_effect_predictor_install.t
```diff
@@ -139,6 +139,18 @@ my %specific_argument = (
         inputs_ref      => [qw{ homo_spaiens }],
         expected_output => q{--SPECIES homo_spaiens},
     },
+    version => {
+        input           => q{91},
```
henrikstranneheim: Why not make it a number and not a string and use READONLY constants instead?

lib/MIP/Program/Variantcalling/Vep.pm
jemten: I think the noupdate should be the default. If you don't give the noupdate option and there is a more recent version of vep available the installation will pause and wait for user input. If you decide that you want the later version you will break the installation of any subsequent programs. It is also more in line with how the other programs handles the version flags.  I'll change it to be on by default and remove the hardcoded on in the subroutine call. 
I don't know if it is meaningful to have it as possible option to supply on the command line or config for the reasons mentioned above.

t/variant_effect_predictor_install.t
jemten: i was a bit lazy

t/variant_effect_predictor_install.t
```diff
@@ -31,6 +31,7 @@ our $VERSION = 1.1.0;
 Readonly my $SPACE   => q{ };
 Readonly my $NEWLINE => qq{\n};
 Readonly my $COMMA   => q{,};
+Readonly my $NUMBER  => 91;
```
henrikstranneheim: NUMBER is pretty anonymous. How about VEP_VERSION?

henrikstranneheim:
👍

PR #472, opened 2018-04-17
get paths for GATK, picard, vep and snpEff, fixes #466

### LINKED ISSUE #466: Set default paths to GATK, picardtools, vep and snpeff
Set a dynamic default path to:
- [ ] GATK
- [ ] Picardtools
- [ ] VEP, VEP cache, VEP plugins
- [ ] SNPEff

jemten:


lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+    }
+    if ( not -d $active_parameter_href->{conda_path} ) {
+        return q{Failed to find default path};
+    }
+    my $conda_path = $active_parameter_href->{conda_path};
+
+    ## Get module and program environments in use
+    my %environment;
+    if (
+        any { $_ eq q{program_source_environment_command} }
```
henrikstranneheim: Can't you just ask: if( exists $active_parameter_href->{program_source_environment_command}).

Does not look like you have to loop through all keys, since you know what you are looking for.

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+        keys %{$active_parameter_href}
+      )
+    {
+        @environment{
```
henrikstranneheim: Add as comment: Build hash with "pprogram_name" as keys and "source env command" as value

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+            keys
+              %{ $active_parameter_href->{program_source_environment_command} }
+          } =
+          values
+          %{ $active_parameter_href->{program_source_environment_command} };
+    }
+    if (
```
henrikstranneheim: Same here! Maybe I am missing something? Why do you loop?

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+        any { $_ eq q{module_source_environment_command} }
+        keys %{$active_parameter_href}
+      )
+    {
+        @environment{
```
henrikstranneheim: Add comment as above

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+    if ( $environment_key and $environment{$environment_key} ) {
+        my @environment_commands = split $SPACE, $environment{$environment_key};
+        $environment = pop @environment_commands;
+        $test_path =
+          catfile( $conda_path, q{envs}, $environment, q{bin}, $bin_file );
+    }
+    ## Check if main environment in use
+    elsif ( $active_parameter_href->{source_main_environment_commands} ) {
+        $environment =
+          @{ $active_parameter_href->{source_main_environment_commands} }
```
henrikstranneheim: Ah, yes it is unfortunate that one is an array and the other a string. We have to standardise that. Could you make an issue so that we do not forget.

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+          } =
+          values %{ $active_parameter_href->{module_source_environment_command}
+          };
+    }
+
+    ## Get environment and set test path;
+    my $environment;
+    my $test_path;
+    ## Check environments
+    if ( $environment_key and $environment{$environment_key} ) {
```
henrikstranneheim: This look like a _get_bin_file_path sub routine to me. That return abs_path($test_path). Et voila you can test the logic (if required) and make a fancy name of what the block does to increase readibility and reduce maintainance, 😄!

lib/MIP/Set/Parameter.pm
```diff
@@ -839,4 +874,144 @@ sub _get_default_repeat_specs_dir_path {
...
+    else {
+        $test_path = catfile( $conda_path, q{bin}, $bin_file );
+    }
+
+    ## Get absolute path
+    $test_path = abs_path($test_path);
+
+    ## Test if path exists
+    if ( not $test_path ) {
+        return q{Failed to find default path};
```
henrikstranneheim: For $bin_file in $NEWLINE, join $NEWLINE, @environment_commands. 
But with proper code
Would be nice to have some more info of what is missing

lib/MIP/Set/Parameter.pm
jemten: True! In my first attempts i tried to test for the dereferenced value ( ${$active_parameter_href->{program_source_environment_command} ) and I got reference errors when the program_source_environment_command wasn't defined in the config.  Forgot that I could test for the reference :)

lib/MIP/Set/Parameter.pm
jemten: See comment above

lib/MIP/Set/Parameter.pm
jemten: I'll sub it! I think that the logic in that part is already tested in the test I added to t/set_custom_default_to_active_parameter.t

henrikstranneheim:
👍

PR #491, opened 2018-05-07
rna base recal, fix #486

### LINKED ISSUE #486: Write RNA recipe for GATK BaseRecalibration
In the rare disease pipeline the samples are processed per contig. In the RNA pipe we will process the whole bam file directly

jemten:


lib/MIP/Recipes/Analysis/Star_fusion.pm
```diff
@@ -204,67 +205,68 @@ sub analysis_star_fusion {
...
-        say {$FILEHANDLE} $AMPERSAND . $SPACE . $NEWLINE;
-    }
-    say {$FILEHANDLE} q{wait}, $NEWLINE;
 
-    ## Close FILEHANDLES
-    close $FILEHANDLE or $log->logcroak(q{Could not close FILEHANDLE});
+        my $process_batches_count = 1;
 
-    if ( $mip_program_mode == 1 ) {
+        while ( my ( $sample_id_index, ) =
```
henrikstranneheim: Why the loop over all samples. Is this not processing per sample_id now?

t/check_vcfanno_toml.t
```diff
@@ -0,0 +1,187 @@
...
+my $log = initiate_logger(
+    {
+        file_path => $test_log_path,
+        log_name  => q{TEST},
+    }
+);
+
+## Given a frequency file and toml file, when matching records
+my $sv_vcfanno_config_file =
+  catfile(qw{ a test dir GRCh37_all_sv_-phase3_v2.2013-05-02-.vcf.gz });
```
henrikstranneheim: I think I have have changed this once again, I presume it will be handled in the merge.

lib/MIP/Recipes/Analysis/Gatk_baserecalibration.pm
```diff
@@ -1130,4 +1133,375 @@ sub analysis_gatk_baserecalibration_rio {
...
+        }
+    );
+    say {$FILEHANDLE} $NEWLINE;
+
+    ## GATK PrintReads
+    say {$FILEHANDLE} q{## GATK PrintReads};
+
+    $outfile_path = $outfile_path_prefix . $outfile_suffix;
+    my $base_quality_score_recalibration_file =
+      $file_path_prefix . $UNDERSCORE . $DOT . q{grp};
```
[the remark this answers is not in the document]
jemten: The underscore shouldn't be there. I will remove it

lib/MIP/Recipes/Analysis/Star_fusion.pm
jemten: I talked to Jesper and we are thinking of changing the way we run STAR-fusion. Now STAR-fusion operates on the junction file from STAR. But STAR operates on read-pair basis eg one junction file per read-pair. This means that some junctions will get to few read to be correctly called and we haven't found a good way of merging the junction files prior to running STAR-fusion. Either we could concatenate the fastq files prior to running STAR and thus only generate one junction file per sample for STAR-fusion. Or we could run STAR-fusion in a mode that also runs STAR and in that chain concatenate the fastq files. However then we would effectively be running STAR twice. Let's decide which strategy to take when you get in this week.

The gist of it is that I will leave STAR-fusion as it is for now and then update the recipe in a separate pull request when we have decided how to run it.

henrikstranneheim:
👍

PR #501, opened 2018-05-18
Recipe for RNA haplotypecaller, resolves #484

### LINKED ISSUE #484: Add Haplotype caller to RNA pipe

jemten:


lib/MIP/Recipes/Analysis/Gatk_haplotypecaller.pm
```diff
@@ -487,4 +489,351 @@ sub analysis_gatk_haplotypecaller {
...
+            outfile_path => $temp_directory,
+        }
+    );
+    say {$FILEHANDLE} q{wait} . $NEWLINE;
+
+    ## GATK HaplotypeCaller
+    say {$FILEHANDLE} q{## GATK HaplotypeCaller};
+
+    my $pcr_indel_model;
+    if ( $active_parameter_href->{gatk_haplotypecaller_pcr_indel_model} ) {
```
henrikstranneheim: Why do you need this if condition? Will the result not be the same with:
my $pcr_indel_model = $active_parameter_href->{gatk_haplotypecaller_pcr_indel_model}

lib/MIP/Recipes/Analysis/Gatk_haplotypecaller.pm
```diff
@@ -487,4 +489,351 @@ sub analysis_gatk_haplotypecaller {
...
+        ## Set to GATK default (CONSERVATIVE) for RNA-seq
```
henrikstranneheim: '##' Should be set to GATK default (CONSERVATIVE) for RNA-seq
Maybe we should add a check either FATAL or WARN that this is so in the sanity check part of MAIN::Analyse

lib/MIP/Recipes/Analysis/Gatk_haplotypecaller.pm
jemten: True we don't need it anymore. It was a remain from the rare disease haplotypecaller recipe

lib/MIP/Recipes/Analysis/Gatk_haplotypecaller.pm
jemten: Possibly. The ## statement I wrote was perhaps more assertive than I ment it to be. CONSERVATIVE is GATK default when running haplotypecaller but I'm not entirely sure that it is the optimal model. Kind of depends on the prep and quality of the RNA I guess. We use NONE for the PCR free wgs libraries

henrikstranneheim:
👍

PR #528, opened 2018-06-25
"Trinity"

J35P312:
Howdy partners!
I compared trinity to cufflinks, and decided to use cufflinks instead (it is faster, and it classifies the transcripts for us, so we wont have to make a long de novo pipeline).

These are command module and test!
Tata!
//J35P312


henrikstranneheim:
👍

PR #575, opened 2018-09-14
Feature/gatk4 variant recalibration

jemten:


jemten:
Travis fails to find TIDDIT in path. The bug will be addressed in a later pull request

PR #607, opened 2018-10-16
Uodated fastqc, added docs on env addition in install

henrikstranneheim:
- Refactored fastqc to avoid migration
- Added install hash to test_mip_hashes fixtures
- Added fastqc in seperate env to avoid perl version clash
- Added general sub in install to get installation parameter attributes. Used it to get full installation envs from parameter file.
- Updated analysis recipe sample to use %io and other minor syntax


lib/MIP/Get/Parameter.pm
```diff
@@ -319,6 +320,57 @@ sub get_dynamic_conda_path {
...
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    ## Make sure that the supplied key exists
+    if ( not exists $parameter_href->{$parameter_name} ) {
+
+        say {*STDERR}
+          qq{Could not find parameter_name key: '$parameter_name' in hash};
+        exit 1;
+    }
```
jemten: Maybe this could be written as: `croak qq{Could not find parameter_name key: '$parameter_name' in hash};`

documentation/Code/Install/Adding_an_env.md
```diff
@@ -0,0 +1,21 @@
...
+    python: 3.6
+  pip:
+  shell:
+
+2. Add the environment name to the `environment_name` hash
+environment_name:
+  e[ENV_NAME]
+3. Add the environment name to the `installations` array
+installations:
+  - e[ENV_NAME]
```
jemten: We should mention that the environment needs to be added to the cli as well

lib/MIP/Get/Parameter.pm
```diff
@@ -319,6 +320,57 @@ sub get_dynamic_conda_path {
     return catdir(@bin_path_dirs);
 }
 
+sub get_install_parameter_attribute {
```
jemten: Nice sub!

lib/MIP/Get/Parameter.pm
henrikstranneheim: I know, but I am having difficulties trapping and then testing the output from croak(). Are you aware of a good way to do it?

documentation/Code/Install/Adding_an_env.md
henrikstranneheim: Good point . I will add it

lib/MIP/Get/Parameter.pm
henrikstranneheim: 😄 Thx

lib/MIP/Get/Parameter.pm
jemten: Did some quick tests and the exit signal can be trapped with `$trap->die` but strangely I can't find a good way to trap the error message. Feels like it should be easy... However, if we've both tried and couldn't find a way I think we should let it rest for now and merge

henrikstranneheim:
@jemten Plz review again.

- Had to rename the efastqc to a more perl oriented env because Bwa, Bwa-kit also needs perl
- Found a way to trap and test croak() 👍

PR #610, opened 2018-10-18
program for running mip code checks

jemten:
Draft for a simple bash script that automates perltidy and perlcritic. As default it runs on the files that are new or have been modified. Perl critic severity has been set to level 2 as default. Increased the code line length to 90 and tried to change the number of lines in between open and close filehandles. 

Later we can include a check to see that the $VERSION line has been updated for modified files. Open for suggestions on how to change it or modify the behaviour of perlcritic and perltidy further (or skip modifications all together)

### DIFF
--- a/.perlcriticrc_mip
+++ b/.perlcriticrc_mip
@@ -1,0 +1,5 @@
+## Modifications to perl critic policies for MIP development
+
+## Increase the number of lines between open and close filhandle, default 9
+[InputOutput::RequireBriefOpen]
+lines = 100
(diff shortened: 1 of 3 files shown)


henrikstranneheim:
Could we have severity 1 and add this to the .perlcriticrc_mip:
```
[InputOutput::RequireCheckedSyscalls]
functions = :builtins
exclude_functions = print say
```

jemten:
Sure! Will add it to the .perlcriticrc_mip file. I'll also add to documentation

henrikstranneheim:
Great! Looks good to me! Will be fun to try

PR #617, opened 2018-10-23
download gencode gtf

jemten:
Download transcript annotation file from Gencode. 
All rd_rna recipes that requires a gtf file now uses the same.


henrikstranneheim:
👍

PR #641, opened 2018-11-16
launch pip as a python module, fixes #637  :zap:

### LINKED ISSUE #637: Add 'python -m' before pip command in install
This ensures we are using the correct python interpreter as conda seems to have trouble switching when using conda activate [env] with conda base (python v3) and activated env (python v 2.7)

Affected recipes: rhocall, vcf2cytosure
https://github.com/pypa/pip/issues/5599

jemten:


henrikstranneheim:
👍

PR #689, opened 2018-12-20
move ggf compare output

jemten:


lib/MIP/Recipes/Analysis/Gffcompare.pm
```diff
@@ -251,20 +253,40 @@ sub analysis_gffcompare {
...
-        {
-            FILEHANDLE   => $FILEHANDLE,
-            infile_path  => $gff_output_path,
-            outfile_path => $outfile_path,
-        }
+
+    my @file_names = (
+        [ $gff_output_path, $outfile_path ],
+        [
+            catfile(
```
henrikstranneheim: Maybe you should assign this to a scalar outside of the array assigment, to make less bloated.

jemten: Yeah, it isn't pretty. Will do that

PR #710, opened 2019-01-22
Feature/sample info

henrikstranneheim:


lib/MIP/Check/Pipeline.pm
```diff
@@ -374,7 +374,7 @@ sub check_rd_dna {
     );
 
     ## Add to sample info
-    add_to_sample_info(
+    set_to_sample_info(
```
jemten: Do we need the to? Feels like set_sample_info clearer is clearer. I'm sensing a regex has been at work here ;)

lib/MIP/QC/Sample_info.pm
```diff
@@ -24,18 +24,18 @@ BEGIN {
...
-      add_recipe_metafile_to_sample_info
-      add_recipe_outfile_to_sample_info
-      add_to_sample_info
+      set_gene_panel
+      set_infile_info
+      set_most_complete_vcf
+      set_parameter_to_sample_info
+      set_processing_metafile_to_sample_info
+      set_recipe_metafile_to_sample_info
+      set_recipe_outfile_to_sample_info
```
jemten: Same with the 'to' in the `set_SOMETHING_to_sample_info`. The to made more sense when it was add. Maybe 'in' or 'for' makes more sense? But this is nitpicking really.

henrikstranneheim: Agreed! Replacing to 'in'. The 'set_in_sample_info' is a bit agnostic, but currently serves as a sub when nothing else fits. We can change that in a future PR.

henrikstranneheim:
I had to rebase due to the VEP_RNA. Closing this PR and reopening another.

PR #715, opened 2019-01-29
fix(RNA-pipe): Fixes to the Rd_rna pipeline

jemten:
- Add analysisrunstatus to RNA pipe
- Fix for running salmon quant on multi lane samples
- Changes some of the paths that are written to qc_sample_info

Will move the single-end/paired-end check in a later pull request

### DIFF
--- a/lib/MIP/Get/Parameter.pm
+++ b/lib/MIP/Get/Parameter.pm
@@ -21,7 +21,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.09;
+    our $VERSION = 1.10;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{
@@ -41,6 +41,7 @@ BEGIN {
       get_read_group
       get_recipe_parameters
       get_recipe_attributes
+      get_sequence_run_type
       get_user_supplied_info
     };
 }
@@ -1031,7 +1032,12 @@ sub get_recipe_attributes {
             store       => \$parameter_href,
             strict_type => 1,
         },
-        recipe_name => { store => \$recipe_name, strict_type => 1, },
+        recipe_name => {
+            required    => 1,
+            defined     => 1,
+            store       => \$recipe_name,
+            strict_type => 1,
+        },
     };
 
     check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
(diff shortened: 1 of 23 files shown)


lib/MIP/Program/Variantcalling/Salmon.pm
```diff
@@ -218,24 +228,28 @@ sub salmon_quant {
...
         q{-1}
       . $SPACE . q{<(}
       . $read_files_command
       . $SPACE
-      . $read_1_fastq_path
+      . join( $SPACE, @{$read_1_fastq_paths_ref} )
       . $SPACE . q{)};
 
-    if ($read_2_fastq_path) {
+    if ($read_2_fastq_paths_ref) {
```
henrikstranneheim: Will this ref pointer always exists? should it be "@{$read_2_fastq_paths_ref}"?

lib/MIP/Recipes/Analysis/Salmon_quant.pm
```diff
@@ -161,204 +159,174 @@ sub analysis_salmon_quant {
...
             }
         )
     );
-    my $outdir_path                = $io{out}{dir_path};
-    my @outfile_name_prefixes      = @{ $io{out}{file_name_prefixes} };
-    my @outfile_paths              = @{ $io{out}{file_paths} };
-    my @temp_outfile_path_prefixes = @{ $io{temp}{file_path_prefixes} };
-    my @temp_outfile_paths         = @{ $io{temp}{file_paths} };
+    my $outdir_path  = $io{out}{dir_path};
+    my $outfile_name = ${ $io{out}{file_names} }[0];
```
henrikstranneheim: $io{out}{file_names}->[0]
$io{out}{file_paths}->[0]
Is better

lib/MIP/Recipes/Analysis/Salmon_quant.pm
```diff
@@ -161,204 +159,174 @@ sub analysis_salmon_quant {
...
-        my $file_path           = $temp_outfile_paths[$infile_index];
-        my $file_path_prefix    = $temp_outfile_path_prefixes[$infile_index];
-        my $outfile_name_prefix = $outfile_name_prefixes[$infile_index];
-        my $outfile_path        = $outfile_paths[$infile_index];
-
-        # Collect paired-end or single-end sequence run mode
-        my $sequence_run_mode =
+    foreach my $infile_prefix (@infile_prefixes) {
+        push @sequence_run_modes,
           $sample_info_href->{sample}{$sample_id}{file}{$infile_prefix}
```
henrikstranneheim: We should write a sub to "get" sequence run modes and use it throughout instead of hard coding the path in the hash. If you do not want to do it in this PR. Please open an issue so that we can do when there is time

lib/MIP/Recipes/Analysis/Salmon_quant.pm
```diff
@@ -161,204 +159,174 @@ sub analysis_salmon_quant {
...
-                FILEHANDLE                      => $FILEHANDLE,
-                job_id_href                     => $job_id_href,
-                log                             => $log,
-                recipe_directory                => $recipe_name,
-                recipe_name                     => $recipe_name,
-                process_time                    => $time,
-                source_environment_commands_ref => \@source_environment_cmds,
-                temp_directory                  => $temp_directory,
-            }
+    # Fail on mixed sequencing modes - THIS CHECK WILL BE MOVED IN A LATER PR
```
henrikstranneheim: Also make an issue - so that we do not forget

lib/MIP/Recipes/Analysis/Salmon_quant.pm
```diff
@@ -161,204 +159,174 @@ sub analysis_salmon_quant {
...
+    ## Salmon quant
+    say {$FILEHANDLE} q{## Quantifying transcripts using } . $recipe_name;
 
-        ### Get parameters
-        ## Infile(s)
-        my @fastq_files = $temp_infile_paths[$paired_end_tracker];
+    ## For paired_end, split first and second reads
+    if ( $sequence_run_mode eq q{paired-end} ) {
+        my @read_1_fastq_paths =
+          @temp_infile_paths[ grep { !( $_ % 2 ) } 0 .. $#temp_infile_paths ];
```
henrikstranneheim: A comment on the grep would be nice

lib/MIP/Recipes/Analysis/Vep.pm
```diff
@@ -1316,6 +1316,7 @@ sub analysis_vep_rna {
         }
     );
 
+    my $outfile_name   = ${ $io{out}{file_names} }[0];
```
henrikstranneheim: Use deref arrow instead

t/salmon_quant.t
```diff
@@ -91,17 +91,20 @@ my %required_argument = (
...
+    outdir_path => {
         input           => catfile(qw{ dir output_dir }),
-        expected_output => q{--output} . $SPACE . catfile(qw{ dir output_dir }),
+        expected_output => q{--output} . $SPACE . catdir(qw{ dir output_dir }),
     },
-    read_1_fastq_path => {
-        input           => catfile(qw{ dir hej_r1.fastq.gz }),
+    read_1_fastq_paths_ref => {
+        inputs_ref =>
+          [ catfile(qw{ dir hej_r1.fastq.gz }), catfile(qw{ dir hello_r1.fastq.gz }) ],
```
henrikstranneheim: Let's keep it one language 😄 . Maybe hi instead of hej

lib/MIP/Get/Parameter.pm
```diff
@@ -1158,4 +1164,74 @@ sub get_user_supplied_info {
     return %user_supply_switch;
 }
 
+sub get_sequence_run_type {
+
```
henrikstranneheim: I would suggest moving this into QC/Sample_info.pm, since we are calculating this based on user input and it only involves the sample info hash. What do you think?

PR #722, opened 2019-02-07
feature(constants): Added a constants.pm for storing constants

henrikstranneheim:
- Added Moose::App version flag
- Export $MIP_VERSION from Constants
- Import $MIP_VERSION from Constants and use with Moose::App version command (mip version)
- Removed mip version from MAIN/Analysis.pm
- Removed utility sub write_script_version and associated test


lib/MIP/Cli/Mip.pm
```diff
@@ -9,14 +10,15 @@ use warnings qw{ FATAL utf8 };
 
 ## CPANM
 use autodie qw{ :all };
-use MooseX::App qw{ BashCompletion Color MutexGroup Typo };
+use MooseX::App qw{ BashCompletion Color MutexGroup Typo Version };
```
jemten: Nice! 
Reading the MooseX documentation it seems like we will be able to have MooseX display the license, author and copyright together with the version. Something we can add later for that extra flair of professionalism :)

henrikstranneheim: 👍

PR #723, opened 2019-02-08
fix(recipe_memory): Recipes don't request more memory than what is av…

jemten:
…ailable


henrikstranneheim:
Just fix the travis tests

PR #741, opened 2019-02-20
fix(installation): master installation on hasta

jemten:
- updated installation of svdb and vcf2cytosure
- added libstdcxx-ng dependeny in order to get the R part of picardtools collectmultiplemetrics to work
- added gatk dependency to cnvnator


henrikstranneheim:
👍

PR #748, opened 2019-02-22
feat(vcf2cytosure): vcf2cytosure_var_size

jemten:
Made vcf2cytosure variant size an option that you can set from command line or config


henrikstranneheim:
👍

PR #759, opened 2019-03-12
feat(bwa_mem): Refactored bwa mem recipe

henrikstranneheim:
- Split bwa mem recipe for "bwa mem" and "run-bwa mem" into 2
  independent recipes
- Depending on genome version either of the recipes will be initilized
- Moved get_read_group to sample_info.pm
- Created sub for getting interleaved status from sample_info hash
- Added use of get_sequencing_run_type to star aln recipe
- Added methods docs in API for Sample_info

### DIFF
--- a/documentation/API/Sample_info.md
+++ b/documentation/API/Sample_info.md
@@ -71,3 +71,58 @@ sample: { #Hash of hashes
   },
 },
 ```
+
+## Methods
+get_read_group:
+Return hash with read group headers.
+```Perl
+my %read_group = get_read_group(
+            {
+                infile_prefix    => $infile_prefix,
+                platform         => $active_parameter_href->{platform},
+                sample_id        => $sample_id,
+                sample_info_href => $sample_info_href,
+            }
+        );
+$rg{id} = <$infile_prefix>;
+$rg{pu} = <flowcell>.<lane>.<sample_barcode>;
+$rg{sm} = <$sample_id>;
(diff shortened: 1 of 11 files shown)


henrikstranneheim:
Ignore the coming Travis fail on temp_dir it was fixed in the previous PR, but I have not updated in this feature branch.

t/analysis_bwa_mem.t
```diff
@@ -67,7 +66,7 @@ diag(   q{Test analysis_bwa_mem from Bwa_mem.pm v}
 my $log = test_log( { log_name => q{MIP}, no_screen => 1, } );
 
 ## Given parameters
```
jemten: The 'given - then' comments in this test offers very little in terms of understanding what's being tested. 
If week keep the syntax this minimal then I don't really see the point of using it. I would suggest adding a few more word.

The same goes for the other test

henrikstranneheim: Agreed. For integration tests the given-when-then syntax is less relevant as we are not really testing many edge cases. I add a few more words basically saying "## Given download parameters for recipe" to make it clear that this is a higher level test, that will have to sufice for now.

henrikstranneheim: Let's move Sample_info in another (tuesday, 😄 ) PR. I also think it is more logically placed straight under MIP.

PR #780, opened 2019-03-31
feature(rna_cleanup): remove cp to temp dir

jemten:
- removed cp to temp dir for rna recipes
- made assembly of RNA transcripts more stringent


jemten:
@henrikstranneheim,
Coverage drops since the subs migrate_files and xargs_migrate_contig_files aren't used anymore. We should either remove them or write tests for them. I can write some tests and attach them to this PR if we want to keep them around. What do you think?

henrikstranneheim:
Actually, let's remove them. I think it is better so start from scratch when (if) the situation arises that we need the functionality again so that is it tailored towards solving that need and not a historical need. Furthermore, dead code has a way of becoming unusable after a while anyway.

PR #922, opened 2019-08-20
feat(test): py3 compability test

jemten:
### This PR fixes:

- Fixes python 3 compatibility test

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Check/Installation.pm
+++ b/lib/MIP/Check/Installation.pm
@@ -23,7 +23,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.04;
+    our $VERSION = 1.05;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{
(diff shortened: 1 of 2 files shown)


lib/MIP/Check/Installation.pm
```diff
@@ -229,7 +229,7 @@ sub check_python_compability {
 
     ## Cover the case where no program has been actively chosen for installation
     my @programs_to_check;
-    if ( not defined $select_programs_ref ) {
+    if ( scalar @{$select_programs_ref} == 0 ) {
```
henrikstranneheim:
```suggestion
    if ( @{$select_programs_ref} ) {
```
Would not this work as well?

jemten: Not a the moment. The suggestion would return true for a defined empty array.

henrikstranneheim: 👍

Issue #95, opened 2016-02-11
Support for hg38 from fastq to ranked variants


henrikstranneheim:
**Recipes**
- [x] fastqc
- [x] bwa_mem
- [x] picardtools_mergesamfiles
- [x] markduplicates
- [x] gatk_baserecalibration
- [x] chanjo_sexcheck
- [x] samtools_subsample_mt
- [x] tiddit_coverage
- [ ] sambamba_depth (test new exons file: "grch38_scout_exons-2019-09-16-.bed";  Clinical-Genomics/scout#644)
- [x] picardtools_collectmultiplemetrics
- [ ] picardtools_collecthsmetrics (test new exons file: "grch38_scout_exons-2019-09-16-.bed";  Clinical-Genomics/scout#644)
- [x] cnvnator_ar
- [x] expansionhunter 3.0.0 with 3.0.1 var_cat_file 
- [x] manta
- [x] tiddit
- [x] delly_call
- [x] delly_reformat
- [x] sv_combinevariantcallsets
- [ ] sv_annotate (works with INFO field reduced gnomad file reset memory to 7Gb)
  Needs liftover files to grch38: 
  - grch37_svdb_query_clingen_cgh_benign_-v1.0.0-.vcf (Daniel: ?)
  - grch37_svdb_query_clingen_cgh_pathogenic_-v1.0.0-.vcf (Daniel: ?)
  - grch37_svdb_query_clingen_ngi_-v1.0.0-.vcf (Daniel: Skip)
  - grch37_mip_sv_svdb_export_-2018-10-09-.vcf (Rerun cases ~200)
  - grch37_swegen_concat_sort_-20170830-.vcf (Waiting: Jesper)
  - grch37_svdb_query_decipher_-v1.0.0-.vcf (Daniel investigates)
  - grch37_gnomad.genomes_-r2.1.1_sv-.vcf (Daniel: Skip)
- [x] sv_varianteffectpredictor
- [x] sv_vcfparser
- [ ] sv_rankvariant
  Need update to grch38:
   - grch37_cust003-cmms-red-pen_-2017-.tsv (Clinical-Genomics/scout#1300)
- [x] sv_reformat
- [ ] vcf2cytosure_ar (dnil comment below)
- [x] gatk_haplotypecaller (but long runtime 21 h,solved with reduced gnomad file)
- [x] gatk_genotypegvcfs
  Need to reprocess for grch38 (WES)
[...]

PR #952, opened 2019-09-03
Feature/grch38 3

henrikstranneheim:
### This PR fixes:

- Fixes grch38 support processing wise from fastq to ranked variants
- Still some reference and files need to be built to make a complete workflow #95 
- Spidex need to be replaced with another splice tool
- Rank models are updated but WIP since the necessary input needs to be built or formally rejected.

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim:
Fixed travis in develop

PR #1001, opened 2019-09-13
Feature/vep plugin hash

henrikstranneheim:
### This PR fixes:

- Vep plugins as hash parameter
- Also adds $GENOME_VERSION and $GENOME_SOURCE for future use

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Check/Parameter.pm
+++ b/lib/MIP/Check/Parameter.pm
@@ -2006,7 +2006,7 @@ q{Could not retrieve VEP cache version. Skipping checking that VEP api and cache
 sub check_vep_plugin {
 
 ## Function : Check VEP plugin options
-## Returns  :
+## Returns  : 0 or 1 
 ## Arguments: $log             => Log object
 ##          : $parameter_name  => Parameter name
 ##          : $vep_plugin_href => VEP plugin annotation {REF}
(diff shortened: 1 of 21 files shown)


lib/MIP/Check/Parameter.pm
```diff
@@ -2002,6 +2003,73 @@ q{Could not retrieve VEP cache version. Skipping checking that VEP api and cache
     return 1;
 }
 
+sub check_vep_plugin {
+
+## Function : Check VEP plugin options
+## Returns  :
```
henrikstranneheim:
```suggestion
## Returns  : 0 or 1 
```

t/check_vep_plugin.t
```diff
@@ -0,0 +1,116 @@
...
+
+my $is_ok = check_vep_plugin(
+    {
+        log             => $log,
+        parameter_name  => q{vep_plugin},
+        vep_plugin_href => \%vep_plugin,
+    }
+);
+
+## Then
```
henrikstranneheim:
```suggestion
## Then return true
```

henrikstranneheim:
True that! I will remove it.

PR #1037, opened 2019-09-30
feat(rank_model): Added current working draft of rank model

henrikstranneheim:
### This PR fixes:

- Adds a current working draft on the new rank model with VEP CSQ fields for CLINVAR and dbnsfp that was previously found in the VCF INFO field.
- Removed some old templates

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


templates/rank_model_cmms_-v1.26-.ini
```diff
@@ -1,5 +1,5 @@
 [Version]
-  version = 1.23
+  version = 1.26
```
jemten: That's quite a jump :)

henrikstranneheim: I actually based on 1.25 and not 1.23 I think git went wrong somewhere

PR #1040, opened 2019-09-30
feat(references): Adds writing a yaml file for all active references …

henrikstranneheim:
…used in a

analysis - for future reference...
- Added test
- Added to all pipelines

### This PR fixes:

- 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim:
But I did -  to the definitions file and and in the pipeline check. I think it should be online. If not let us fix it in another PR.

PR #1056, opened 2019-10-07
Feature/chanjo sif

henrikstranneheim:
### This PR fixes:

- Adds chanjo with sambamba sif
- Adds sambamba standalone sif
- Removes sambamba shell installation

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass


containers/Singularity.chanjo-4.2.0
```diff
@@ -0,0 +1,59 @@
...
+    conda config --add channels conda-forge
+    conda config --add channels bioconda
+
+    ## Install conda packages
+    conda install --yes sambamba=0.6.8=h682856c_0 pip python=3.7 ldc=1.11.0=hb2c9046_0
+
+    ## Clean up after conda
+    /opt/conda/bin/conda clean -tipsy
+
+    pip install chanjo==4.2.0
```
jemten:
```suggestion
    pip install --no-cache-dir chanjo==4.2.0
```

jemten: There is no reason to save to cache since we won't install anything else in the image once it is built. Not having a cache will make the image marginally smaller. This is not part of the template but we could consider adding it.

PR #1059, opened 2019-10-08
Feature/bam contigs

henrikstranneheim:
### This PR fixes:

- Adds BAM level contigs set to always include Y even if all samples are female

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


templates/mip_rd_dna_config.yaml
```diff
@@ -14,6 +14,7 @@ load_env:
     installation: emip
     method: conda
     mip:
+    multiqc_ar:
```
jemten:
```suggestion
```
Shouldn't this be unnecessary since MIP will default to the main environment?

PR #1065, opened 2019-10-10
Feature/chromograph

jemten:
### This PR fixes:
- unifies the previous chromgraph commands to one
- upd runs on sample level instead of case
- adds check for possible trios
- upd runs for all possible trios in a case
- the chromograph chain only runs for the rd_dna pipeline

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


templates/mip_rd_dna_config.yaml
```diff
@@ -25,7 +25,7 @@ load_env:
     installation: eperl5
     method: conda
   mip_travis_py3:
-    chanjo_sexcheck:
+      #chanjo_sexcheck:
```
henrikstranneheim:
```suggestion
```

PR #1089, opened 2019-10-14
feat(trim_galore): updates trim_galore, adds sif

jemten:
### This PR fixes:

- see above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Get/Executable.pm
+++ b/lib/MIP/Get/Executable.pm
@@ -23,7 +23,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.00;
+    our $VERSION = 1.01;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{ get_binary_version get_executable };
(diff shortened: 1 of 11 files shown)


lib/MIP/Get/Executable.pm
```diff
@@ -241,6 +241,11 @@ q?'my ($version) = /\(htslib\)\s+(\S+)/xms; if($version) {print $version;last;}'
             version_regexp =>
               q?'my ($version) = /TIDDIT-(\S+)/xms; if($version) {print $version;last;}'?,
         },
+        trim_galore => {
```
henrikstranneheim: Nice!

henrikstranneheim: Maybe add "pigz" as well?

jemten: Good idea

jemten: No that I think about it. Should we log the version of these helper programs? It's a bit like logging the version of bzip2

henrikstranneheim: I would not say that it is super critical. If it easy - consider doing it. I am for instance going to skip this with bwakit as it does not output a version in any way.

PR #1096, opened 2019-10-15
Feature/rseqc sif

jemten:
### This PR fixes:

 - adds sif for rseqc
 - adds capture regex for version
 - removes perl5 env for rna

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Check/Installation.pm
+++ b/lib/MIP/Check/Installation.pm
@@ -23,7 +23,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.07;
+    our $VERSION = 1.08;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{
@@ -280,11 +280,9 @@ sub check_python_compability {
             store    => \$log,
         },
         python3_programs_ref => {
-            default     => [],
-            defined     => 1,
-            required    => 1,
-            store       => \$python3_programs_ref,
-            strict_type => 1,
+            default  => [],
+            required => 1,
+            store    => \$python3_programs_ref,
         },
         python_version => {
             required => 1,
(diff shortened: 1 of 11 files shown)


lib/MIP/Check/Installation.pm
```diff
@@ -301,12 +299,15 @@ sub check_python_compability {
 
     use Array::Utils qw{ intersect unique };
 
+    ## Return if no python 3 programs are set
+    return 1 if ( not $python3_programs_ref );
```
henrikstranneheim:
```suggestion
    return 1 if ( not @{$python3_programs_ref} );
```

henrikstranneheim: Otherwise might it not give you an ARRAY ref pointer back?

templates/mip_download_rd_rna_config_-1.0-.yaml
```diff
@@ -1,5 +1,5 @@
 load_env:
-  MIP_rd_rna:
+  D_mip7-rna_AJ:
```
henrikstranneheim: Is it time for that CLI option?

jemten: Yeah 😅

lib/MIP/Check/Installation.pm
jemten: For the RNA installation the array (or even the reference to the array) doesn't exist. I first had it as you wright it but got some error about not being allowed to use an undefined value as an array reference (or something similar). I can create the reference and then check if it point to an array?

henrikstranneheim: No, but then it is fine. You also cover this case in the tests. 👍

templates/mip_download_rd_rna_config_-1.0-.yaml
jemten:
```suggestion
 MIP_rd_rna:
```

PR #1112, opened 2019-10-18
Feature/htslib sif

henrikstranneheim:
### This PR fixes:

- htslib as sif

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass

### DIFF
--- a/containers/Singularity.htslib-1.9
+++ b/containers/Singularity.htslib-1.9
@@ -1,0 +1,58 @@
+Bootstrap: docker
+From: ubuntu:xenial
+
+%help
+    This is a container for installing bcftools, bgzip, htslib, samtools and tabix
+    version 1.9
+
+%labels
+    Maintainer Clinical-Genomics/MIP
+    Version 0.1
+
+%environment
+    SHELL=/bin/bash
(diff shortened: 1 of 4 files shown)


containers/Singularity.htslib-1.9
```diff
@@ -0,0 +1,58 @@
+Bootstrap: docker
+From: ubuntu:xenial
+
+%help
+    This is a container for installing bcftools, bgzip, htslib, samtools and tabix
```
jemten: Nice to bundle the tools! 👍

henrikstranneheim: Yeah, they seem to belong together

henrikstranneheim:
Sorry I missed the comment about the CLI. What did you mean?

jemten:
In the install cli you have the option to select which programs that are to be installed. I think that bcftools and samtools are available options but if you supply them now it won't work. Better to remove them and only have htslib which will install all of them. 
But no worries, we can take that later

henrikstranneheim:
Ah, yes of course now I see what you mean. I will do another PR

PR #1114, opened 2019-10-18
feat(conda): Removed unused conda packages

henrikstranneheim:
### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


jemten:
Also travis fails. 
A quick solution is to let it install something (say python 3). Otherwise we need to rewrite the installation a bit so that a conda env is created even if nothing is to be installed into it

henrikstranneheim:
Agreed! Let's go with python 3. I'll fix

jemten:
> Agreed! Let's go with python 3. I'll fix

👍

PR #1136, opened 2019-11-02
Feature/sif relative path

henrikstranneheim:
### This PR fixes:

- Makes sifs relative to proxy bin

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Recipes/Install/Singularity.pm
+++ b/lib/MIP/Recipes/Install/Singularity.pm
@@ -34,7 +34,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.05;
+    our $VERSION = 1.06;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{ install_singularity_containers };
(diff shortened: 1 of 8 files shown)


lib/MIP/Recipes/Install/Singularity.pm
```diff
@@ -151,6 +152,11 @@ sub install_singularity_containers {
         $container_href->{program_bind_paths} = [];
 
         my $container_path = catfile( $container_dir_path, $container . q{.sif} );
```
jemten: Still needed?

henrikstranneheim: Yes, I think so to download the sif to the selected conda env by the user in mip install. Then we make the relative link. 
It looks like this:
```
singularity pull --force /home/henrik.stranneheim/miniconda3/envs/D_mip8.0_191019_hs/share/containers/cnvnator.sif
```
and
```
echo -n "singularity exec \"\$CONDA_ENV_DIR\"/share/containers/cnvnator.sif
```

henrikstranneheim:
About the CLI - you are quite right I will remove the CLI option and lock this to a specific conda env to enable portability.

lib/MIP/Recipes/Install/Singularity.pm
jemten: Don't you need/want to remove the `if ( not $container_dir_path ) {` above and have container_dir_path to always be `catdir( $conda_env_path, qw{ share containers } )`?

henrikstranneheim: yes, you are right now that it is constant there is no need for the if condition.

henrikstranneheim:
@jemten How about now?

PR #1159, opened 2019-11-14
Feature/vep directory path

henrikstranneheim:
### This PR fixes:

- 

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass

### DIFF
--- a/lib/MIP/Check/Parameter.pm
+++ b/lib/MIP/Check/Parameter.pm
@@ -59,7 +59,7 @@ BEGIN {
       check_sample_id_in_parameter_value
       check_select_file_contigs
       check_vep_custom_annotation
-      check_vep_directories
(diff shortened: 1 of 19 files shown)


lib/MIP/Check/Parameter.pm
```diff
@@ -1848,29 +1848,24 @@ sub check_vep_custom_annotation {
 
 sub check_vep_directories {
```
jemten: Maybe we should rename the sub to something else. We could call it simply check_vep

henrikstranneheim: True that! Decided to be slightly more verbose: check_vep_api_cache_versions

jemten: 👍

PR #1314, opened 2020-01-23
feat(rna_ref): adds rna ref genomes

jemten:
### This PR fixes:

- adds the reference genome parameters for star and salmon

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/templates/mip_download_rd_rna_config_-1.0-.yaml
+++ b/templates/mip_download_rd_rna_config_-1.0-.yaml
@@ -61,17 +61,11 @@ reference_feature:
       decoy_5:
         file: human_g1k_v37_decoy.fasta.gz
         file_check: human_g1k_v37_decoy.fasta.gz.md5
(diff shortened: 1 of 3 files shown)


templates/mip_download_rd_rna_config_-1.0-.yaml
```diff
@@ -61,17 +61,11 @@ reference_feature:
       decoy_5:
         file: human_g1k_v37_decoy.fasta.gz
         file_check: human_g1k_v37_decoy.fasta.gz.md5
-        file_index: human_g1k_v37_decoy.fasta.fai.gz
```
henrikstranneheim: Shoukld this be done for DNA as well?

jemten: I didn't change it since you hadn't experienced any issues with it. I can do it though. Shouldn't be any harm in it

PR #1333, opened 2020-02-04
feat(yamllint): Linting of yaml files

jemten:
### This PR fixes:

  - Adds yamllint to mip-check
  - Linting yaml files in definitions and templates

yamlint needs to be installed. It's available through conda, pip and a multitude of other channels

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


jemten:
We can further configure the linting with the yamllint config file if necessary. This is just the options that I had since before

definitions/not_required_parameters.yaml
```diff
@@ -31,7 +32,6 @@ outfile_suffix:
   key_data_type: SCALAR
 program_executables:
   key_data_type: ARRAY
-  key_data_type: SCALAR
```
henrikstranneheim: ⭐️ NIce catch!

henrikstranneheim:
Perhaps we should supply a bash script that will install perltidy, perl critic, yamllint, Readonly::XS, Yaml::Xs etc in a MIP a development conda env. Then we can also set the versions for each of these and set the standard across developer. What do you think @jemten

mip_install_perl.sh
```diff
@@ -67,12 +73,19 @@ fi
 
 conda install --name "$ENV_NAME" --yes -c bioconda -c conda-forge perl=5.26 perl-app-cpanminus perl-log-log4perl perl-moosex-app perl-file-copy-recursive perl-timedate perl-set-intervaltree
 
-## Source conda
-source "$CONDA_PATH"/etc/profile.d/conda.sh
-
 conda activate "$ENV_NAME"
 
-cpanm --installdeps .
+cd definitions && cpanm --installdeps .
```
henrikstranneheim: Move this to the main level

PR #1346, opened 2020-02-07
feat(display_name): Added use of display name to MIP

henrikstranneheim:
- Get sample_display_name from pedigree
- Rename sample_id in vcf2cytosure input VCF to sample_display_name

### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
+++ b/lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
@@ -171,7 +171,8 @@ sub analysis_vcf2cytosure {
             recipe_name           => $recipe_name,
         }
     );
-    my $ignore_sample_display_name = $active_parameter_href->{ignore_sample_display_name};
+    my $use_sample_id_as_display_name =
+      $active_parameter_href->{vcf2cytosure_use_sample_id_as_display_name};
 
     ## Set and get the io files per chain, id and stream
     my %io = parse_io_outfiles(
@@ -344,9 +345,7 @@ sub analysis_vcf2cytosure {
     say {$filehandle} q{wait}, $NEWLINE;
 
   SAMPLE_ID:
-    while ( my ( $sample_id_index, $sample_id ) =
-        each @{ $active_parameter_href->{sample_ids} } )
-    {
+    foreach my $sample_id ( @{ $active_parameter_href->{sample_ids} } ) {
 
         ## Rename vcf sample
         my $sample_display_name = get_pedigree_sample_id_attributes(
@@ -357,7 +356,7 @@ sub analysis_vcf2cytosure {
             }
         );
 
-        if ( $sample_display_name and not $ignore_sample_display_name ) {
+        if ( $sample_display_name and not $use_sample_id_as_display_name ) {
 
             my $sample_display_vcf_file = catfile( $outdir_path, $sample_display_name );
             $vcf2cytosure_file_info{$sample_id}{in}{$sample_display_name} =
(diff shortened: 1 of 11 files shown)


lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
```diff
@@ -340,6 +343,48 @@ sub analysis_vcf2cytosure {
...
+        ## Rename vcf sample
+        my $sample_display_name = get_pedigree_sample_id_attributes(
+            {
+                attribute        => q{sample_display_name},
+                sample_id        => $sample_id,
+                sample_info_href => $sample_info_href,
+            }
+        );
+
+        if ( $sample_display_name and not $ignore_sample_display_name ) {
```
jemten: I would change the name of the parameter ignore_sample_display_name to something without a negation and default it to 1. Then we can avoid this double negative "and not $ignore_..."and this could be something like:
```Perl 
if ( $sample_display_name and $set_sample_display_name )
```
, which I find more readable.

lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
```diff
@@ -340,6 +343,48 @@ sub analysis_vcf2cytosure {
     }
     say {$filehandle} q{wait}, $NEWLINE;
 
+  SAMPLE_ID:
+    while ( my ( $sample_id_index, $sample_id ) =
```
jemten: Couldn't this be a foreach loop instead or are you using the index for something?

t/analysis_vcf2cytosure.t
```diff
@@ -90,7 +90,7 @@ my %file_info = test_mip_hashes(
     }
 );
 
-## Special case since delly_reformat needs to collect from recipe not immediate upstream
+## Special case since vcf2cytosure needs to collect from recipe not immediate upstream
```
jemten: good catch :)

t/analysis_vcf2cytosure.t
```diff
@@ -131,9 +131,18 @@ $parameter{sv_annotate}{chain}            = q{TEST};
...
+            sex                 => q{male},
+        },
+        ADM1059A2 => {
+            sample_display_name => q{stefan},
+            sex                 => q{unknown},
+        },
+        ADM1059A3 => {
+            sample_display_name => q{venus},
+            sex                 => q{female},
+        },
```
jemten: 🎾 😄

lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
henrikstranneheim: I started with something similar, but then if it is a Boolean with default 1 - how do you turn it off on the command line?

What do you think of: use_sample_iid_as_display_name with default 0

t/analysis_vcf2cytosure.t
henrikstranneheim: 👍! You are in the hall of fame for sure!! 🏆

lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
henrikstranneheim: No, you are right! I Friday afternoon copy of the previous section. I will change it

henrikstranneheim: fixed

lib/MIP/Recipes/Analysis/Vcf2cytosure.pm
jemten: Ah forgot that you need to be able to set it at the command line. It's quite long but I guess it needs to in order to be descriptive

henrikstranneheim: Made it the option recipe specific bit it's long `vcf2cytosure_use_sample_id_as_display_name`. Will have to do for now.

PR #1354, opened 2020-02-11
Feature/blobfish docker

jemten:
### This PR fixes:

- see above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


templates/mip_install_rd_rna_config_-1.0-.yaml
```diff
@@ -41,6 +28,10 @@ singularity:
       arriba: /arriba_v1.1.0/arriba
       draw_fusions.R: /arriba_v1.1.0/draw_fusions.R
     uri: docker://uhrigs/arriba:1.1.0
+  blobfish:
+    executable:
+      BlobFish.py:
```
henrikstranneheim: What about: BlobFish.R

jemten: Thats also an executable in Blobfish, which we never use (I think that it is used internally in Blobfish.py though). That is why I didn't link it

henrikstranneheim: 👍

jemten:
i'll fix the coverage. just have to delete the old install script for blobfish

henrikstranneheim:
👍

PR #1361, opened 2020-02-12
feat(modify_check): Modified sub, renamed and moved check_executable_in_path

henrikstranneheim:
### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass


lib/MIP/Active_parameter.pm
```diff
@@ -347,6 +349,83 @@ sub get_user_supplied_pedigree_parameter {
...
+  PARAMETER:
+    foreach my $parameter_name ( keys %{$active_parameter_href} ) {
+
+        ## Only check path(s) for parameters with "type" key
+        next PARAMETER
+          if ( not exists $parameter_href->{$parameter_name}{type} );
+
+        ## Only check path(s) for parameters with type value eq "recipe"
+        next PARAMETER
+          if ( not $parameter_href->{$parameter_name}{type} eq q{recipe} );
```
jemten: We could potentially change this to a single line with Data::Diver
something like:  
```Perl 
next PARAMETER if ( not Dive( $parameter_href, ( $parameter_name, qw{ type recipe } ) ) )
```
I think that we use Data::Diver in some other places where we want to avoid autovivification, albeit on deeper hashes. 
The original code is fine though and I think there is some merit on not using cpan modules when it is not necessary.

lib/MIP/Active_parameter.pm
```diff
@@ -347,6 +349,83 @@ sub get_user_supplied_pedigree_parameter {
...
+
+        ## Only check path(s) for active recipes
+        next PARAMETER if ( not $active_parameter_href->{$parameter_name} );
```
jemten: Won't this always be true since we are looping over the keys in $active_parameter_href?

lib/MIP/Active_parameter.pm
```diff
@@ -347,6 +349,83 @@ sub get_user_supplied_pedigree_parameter {
...
+
+        ## Alias
+        my $program_executables_ref =
+          \@{ $parameter_href->{$parameter_name}{program_executables} };
```
jemten: Won't this give a warning or something for recipes that lacks executables now that we removed grep and sed?

lib/MIP/Active_parameter.pm
henrikstranneheim: No, I do not think so, since they can also be set to zero.

lib/MIP/Active_parameter.pm
henrikstranneheim: Good suggestion, but this is fine for this example since type is mandatory at that level.

lib/MIP/Active_parameter.pm
henrikstranneheim: I think it will just point to an empty list and then the for loop should not execute

lib/MIP/Active_parameter.pm
jemten: I meant to replace both the if statements with this single one. But as you say this works as well. Sometimes it is good with some extra lines of code if it improves readability 👍

lib/MIP/Active_parameter.pm
jemten: 👍

lib/MIP/Active_parameter.pm
jemten: 👍

lib/MIP/Active_parameter.pm
henrikstranneheim: Ah, I see what you mean. I agree, let's keep it as it is and not introduce the extra deependency and it also reads well

PR #1363, opened 2020-02-13
feat(bcftools_mpileup): Turned it off by default

henrikstranneheim:
### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim:
Yeah, only in the config for production and stage (forgot to change the default), but from now it is by default 😄

PR #1380, opened 2020-02-24
feat(check_recipe_name): Refactor and move sub

henrikstranneheim:
### This PR fixes:

- See above 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Parameter.pm
+++ b/lib/MIP/Parameter.pm
@@ -26,11 +26,12 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.12;
+    our $VERSION = 1.13;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK = qw{
       check_parameter_hash
+      check_recipe_vs_binary_name
       get_cache
       get_capture_kit
       get_order_of_parameters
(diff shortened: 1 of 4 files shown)


lib/MIP/Parameter.pm
```diff
@@ -122,6 +123,60 @@ sub check_parameter_hash {
...
+  RECIPE:
+    foreach my $recipe ( @{$recipe_names_ref} ) {
+
+        next RECIPE if ( not exists $parameter_href->{$recipe}{program_executables} );
+
+      BINARY:
+        foreach my $binary ( @{ $parameter_href->{$recipe}{program_executables} } ) {
+
+            $binary_name{$binary} = undef;
+        }
```
jemten: There's more than one way to do it, but couldn't this be written as:
```Perl
@binary_name{ @{ $parameter_href->{$recipe}{program_executables} } = undef;
```
to save some much valued lines of code ;)

lib/MIP/Parameter.pm
```diff
@@ -122,6 +123,60 @@ sub check_parameter_hash {
...
+
+        next RECIPE if ( not exists $binary_name{$recipe} );
```
jemten: Isn't it better to check the binary name hash for matching recipe name after it has been fully populated? We could potentially have a recipe named after an executable that hasn't been added to the hash yet.

lib/MIP/Parameter.pm
henrikstranneheim: You are right, although it looks a bit heavy for a oneliner. However, but combining your two excellent comments - created another solution 😄. I do not think I managed to save lines, but the functionality is extended and easy to follow (and probably faster...)

lib/MIP/Parameter.pm
henrikstranneheim: I pondered the same thing when I refactored this one. Let´s do it.

henrikstranneheim:
What do you think of this proposed solution?

jemten:
Looks good! The get_cache sub certainly makes it more readable than my somewhat heavy handed oneliner 😄

PR #1387, opened 2020-02-29
feat(set_contigs): Refactor

henrikstranneheim:
- Renamed contigs being set in set_dict_contigs to dict_contigs
- Fixed alt_loci set not being set properly
- Use contigs as controller for setting contigs in file_info
- Made atomic subs for setting contigs

### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass


lib/MIP/File_info.pm
```diff
@@ -24,10 +24,108 @@ BEGIN {
...
+## Function : Set alternative loci contigs
+## Returns  :
+## Arguments: $alt_contig_set_name => Alt contig set identifier
+##          : $file_info_href      => File info hash {REF}
+##          : $primary_contig_href => Primary contig hash {REF}
+
+    my ($arg_href) = @_;
+
+    ## Flatten argument(s)
+    my $alt_contig_set_name;
```
jemten: Is it meaningful to have this as a variable name and not use a fixed key. The way the sub is written now, won't  it will always be the alternative contigs. Maybe if we rewrite the sub to be a more generic contig slicer.

lib/MIP/File_info.pm
```diff
@@ -24,10 +24,108 @@ BEGIN {
...
+
+    my ($arg_href) = @_;
+
+    ## Flatten argument(s)
+    my $bam_contig_set_name;
+    my $file_info_href;
+    my $primary_contigs_ref;
+
+    my $tmpl = {
+        bam_contig_set_name => {
```
jemten: Basically the same comment as above. Though, I guess it could be some benefit to be able to name your bam contig set. You do as you like :)

lib/MIP/File_info.pm
```diff
@@ -204,4 +302,50 @@ sub set_human_genome_reference_features {
...
+## Arguments: $file_info_href          => File info hash {REF}
+##          : $primary_contigs_ref     => Primary contig hash {REF}
+##          : $primary_contig_set_name => Primary contig set identifier
+
+    my ($arg_href) = @_;
+
+    ## Flatten argument(s)
+    my $file_info_href;
+    my $primary_contigs_ref;
+    my $primary_contig_set_name;
```
jemten: see above :)

lib/MIP/Set/Contigs.pm
```diff
@@ -58,34 +58,55 @@ sub set_contigs {
...
+        }
+    );
 
-    ## Contigs sets to set for primary assembly
+    ## Set contigs sets for primary assembly
     my @primary_contig_sets = qw{ contigs contigs_size_ordered };
-    my %bam_contig_set      = (
+
+  PRIMARY_CONTIG_SET:
+    foreach my $contig_set (@primary_contig_sets) {
```
jemten: Now I see...

henrikstranneheim: Yes, the idea was to let the control be handled entirely by the contig module. I thought it would be more clear if all naming happens in the same place. We can discuss it on Monday. I also need to move the contig.pm up one level.

jemten: 👍

PR #1408, opened 2020-03-12
Feature/download envn

henrikstranneheim:
### This PR fixes:

- Added CLI option to rename load_env environment in download template config 
- Lower cased default load_env environments in template config for install and modfied them to be snake_case
- Added r2.1.1_sv download to templates config again
- Fixed bug in gnomad recipe by also including gnomad_sv keep keys
- Fixed bug in decompression by adding newline after sub call

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Active_parameter.pm
+++ b/lib/MIP/Active_parameter.pm
@@ -55,6 +55,7 @@ BEGIN {
       set_default_vcfparser_select_file
       set_exome_target_bed
       set_gender_sample_ids
+      set_load_env_environment
       set_parameter_reference_dir_path
       set_pedigree_sample_id_parameter
       set_recipe_resource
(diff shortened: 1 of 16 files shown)


lib/MIP/Active_parameter.pm
```diff
@@ -1174,6 +1175,53 @@ sub set_default_parameter {
...
+            strict_type => 1,
+        },
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments};
+
+    ## Unpack
+    my $env_name = $active_parameter_href->{environment_name};
+
+    if ( keys %{ $active_parameter_href->{load_env} } != 1 ) {
```
jemten: when would this happen?

jemten: ah! never mind. You're keeping the possibility to use multiple env right?

henrikstranneheim: Exactly. Just preserving existing functionality until we decide to remove it

PR #1421, opened 2020-03-17
feat(refactor):

henrikstranneheim:
### This PR fixes:

- Refactor check_object_suffixes_to_build

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim:
It is because we do not know if it is a "file" or a "directory" we are checking using check_filesystem_objects_existance from the subs point of view. But I agree that this is only relevant to that sub and not to parse_meta_file_suffixes. Will rename "object_suffixes_ref" to something more appropriate

PR #1444, opened 2020-03-27
feat(refactor): check_vep_custom_annotation

henrikstranneheim:
### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Vep.pm
+++ b/lib/MIP/Vep.pm
@@ -27,7 +27,10 @@ BEGIN {
     our $VERSION = 1.00;
 
     # Functions and variables which can be optionally exported
-    our @EXPORT_OK = qw{ check_vep_api_cache_versions };
+    our @EXPORT_OK = qw{
+      check_vep_api_cache_versions
+      check_vep_custom_annotation
+    };
 }
 
 sub check_vep_api_cache_versions {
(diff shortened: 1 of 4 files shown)


lib/MIP/Vep.pm
```diff
@@ -164,4 +222,91 @@ sub _get_vep_cache_species_dir_path {
...
+        },
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    ## Retrieve logger object
+    my $log = Log::Log4perl->get_logger($LOG_NAME);
+
+    ## Check required keys
+    my @required_options = (qw{ key });
```
jemten: Do we expect to have more required keys later?

lib/MIP/Vep.pm
```diff
@@ -164,4 +222,91 @@ sub _get_vep_cache_species_dir_path {
...
+  REQ_OPTION:
+    foreach my $required_option (@required_options) {
+
+        if (   not exists $custom_ann_option_href->{$required_option}
+            or not defined $custom_ann_option_href->{$required_option} )
```
jemten: You sure it shouldn't be 'and' instead of 'or'?

lib/MIP/Vep.pm
henrikstranneheim: catering for the future

Issue #1424, opened 2020-03-19
MIP analysis of small panels 


jemten:
MIP in its current form isn't designed to run small panels and will often fail on such samples. This is somewhat due to a lack of variants in the panels, causing some files to be empty. The solution is to write a new MIP pipeline tailored towards panels.

jemten:
# Output from rtg

## Case: deepalien (NA24143)
Panel: GMCKSolidv4.bed
Read pairs: 118,168,226
```
Threshold  True-pos-baseline  True-pos-call  False-pos  False-neg  Precision  Sensitivity  F-measure
----------------------------------------------------------------------------------------------------
   99.000               4557           4564        183         43     0.9614       0.9907     0.9758
     None               4557           4564        186         43     0.9608       0.9907     0.9755
```

## Case: deepalien (NA24143)
Panel: GMCKSolidv4.bed
Read pairs: 68,648,554
```
Threshold  True-pos-baseline  True-pos-call  False-pos  False-neg  Precision  Sensitivity  F-measure
----------------------------------------------------------------------------------------------------
   99.000               4552           4559        205         48     0.9570       0.9896     0.9730
     None               4553           4560        210         47     0.9560       0.9898     0.9726
````

## Case: sureemu (NA24143)
Panel: GMCKSolidv4.bed
Read pairs: 40,585,090
```
Threshold  True-pos-baseline  True-pos-call  False-pos  False-neg  Precision  Sensitivity  F-measure
----------------------------------------------------------------------------------------------------
   99.000               4554           4560        193         46     0.9594       0.9900     0.9745
     None               4554           4560        200         46     0.9580       0.9900     0.9737
```

PR #1449, opened 2020-03-30
feat(vep): run vep in panel pipe

jemten:
### This PR fixes:

- modifies rna version of vep recipe to use different padding for mt
- use vep rna recipe in panel pipe
- part of #1424 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Recipes/Analysis/Vep.pm
+++ b/lib/MIP/Recipes/Analysis/Vep.pm
@@ -20,7 +20,7 @@ use Readonly;
 
 ## MIPs lib/
 use MIP::Constants
-  qw{ %ANALYSIS $ASTERISK $COMMA $DOT $EMPTY_STR $LOG_NAME $NEWLINE $SPACE $UNDERSCORE };
+  qw{ %ANALYSIS $ASTERISK $COMMA $DOT $EMPTY_STR $LOG_NAME $NEWLINE %PRIMARY_CONTIG $SPACE $UNDERSCORE };
 use MIP::File::Format::Vep qw{ create_vep_synonyms_file };
 
 BEGIN {
@@ -29,7 +29,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.28;
+    our $VERSION = 1.29;
 
     # Functions and variables which can be optionally exported
     our @EXPORT_OK =
@@ -1301,7 +1301,9 @@ sub analysis_vep_rna {
     my $infile_path        = $io{in}{file_path};
     my $infile_suffix      = $io{in}{file_suffix};
 
-    my $job_id_chain = get_recipe_attributes(
+    my $consensus_analysis_type  = $parameter_href->{cache}{consensus_analysis_type};
+    my $genome_reference_version = $file_info_href->{human_genome_reference_version};
+    my $job_id_chain             = get_recipe_attributes(
         {
             parameter_href => $parameter_href,
             recipe_name    => $recipe_name,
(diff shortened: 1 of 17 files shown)


lib/MIP/Recipes/Analysis/Vep.pm
```diff
@@ -1360,14 +1361,32 @@ sub analysis_vep_rna {
...
+    my $vep_synonyms_file_path = create_vep_synonyms_file(
+        {
+            log          => $log,
+            outfile_path => catfile( $outdir_path_prefix, q{synonyms.tsv} ),
+            version      => $genome_reference_version,
+        }
+    );
+
+    ## Get contigs
+    my %primary_contig = Readonly::Clone %PRIMARY_CONTIG;
```
henrikstranneheim: I think this section should be 2 get subs in lib/MIP/Contigs.pm instead. One which returns a specific contig(s) and one that returns all but the specifc contig(s). Then this recipe does not have to know anything about %PRIMARY_CONTIG and its structure and you get 2 general and unit tested sub for reuse elsewhere.

lib/MIP/Recipes/Analysis/Vep.pm
```diff
@@ -1401,35 +1428,67 @@ sub analysis_vep_rna {
...
             vep_features_ref       => \@vep_features_ref,
         }
     );
     say {$filehandle} $NEWLINE;
 
+    ## VEP for MT
+    variant_effect_predictor(
+        {
+            assembly               => $assembly_version,
+            buffer_size            => 20_000,
```
henrikstranneheim: Make it into a READONLY constant

lib/MIP/Recipes/Pipeline/Analyse_rd_dna_panel.pm
```diff
@@ -178,8 +178,9 @@ sub pipeline_analyse_rd_dna_panel {
     use MIP::Recipes::Analysis::Sambamba_depth qw{ analysis_sambamba_depth };
     use MIP::Recipes::Analysis::Samtools_merge qw{ analysis_samtools_merge_panel };
     use MIP::Recipes::Analysis::Variant_integrity qw{ analysis_variant_integrity };
+    use MIP::Recipes::Analysis::Vep qw{ analysis_vep_rna };
```
henrikstranneheim: This intuitively feels strange with `rna` in the Analyse_rd_dna_panel. Could we rename the analysis recipe to something more generic? I know that we usually use pipeline_type in the name, but it feels odd. Maybe we could switch the:
analysis_vep -> analysis_vep_wgs (or analysis_vep_per_contig)
analysis_vep_rna -> analysis_vep
What do you think?

henrikstranneheim:
Seems we lost travis??

henrikstranneheim:
Coverage needs attention as well, 😄

lib/MIP/Recipes/Pipeline/Analyse_rd_dna_panel.pm
jemten: That bugged me as well, but I couldn't come up with a good name for the sub. Will change it for now to: 
analysis_vep -> analysis_vep_wgs
analysis_vep_rna -> analysis_vep

lib/MIP/Recipes/Analysis/Vep.pm
```diff
@@ -1532,6 +1520,8 @@ sub _get_custom_annotation_cmds {
 
     check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
 
+    return if ( not defined $vep_custom_annotation_href );
```
henrikstranneheim: Nice

jemten: thanks!

PR #1462, opened 2020-04-08
feat(refactor): write_references

henrikstranneheim:
### This PR fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim:
Yes, I also considered Analysis.pm since it is the references for a particular analysis, but I think it is better to place the code where the data is located. Then it is more clear where the source is. But as you say some of them are hard to place in the correct context.

PR #1476, opened 2020-04-16
feat(download): small fixes

jemten:
### This PR fixes:

- removed environment name default as it took precedence over config value
- Modified download to bind reference dir to singularity containers

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


jemten:
Ah! Forgot about that

PR #1488, opened 2020-04-23
Feature/update dynamic parameter

jemten:
### This PR fixes:

- the sub update_dynamic_config_parameters can now update arrays

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Update/Parameters.pm
+++ b/lib/MIP/Update/Parameters.pm
@@ -22,7 +22,8 @@ BEGIN {
     our $VERSION = 1.09;
 
     # Functions and variables which can be optionally exported
-    our @EXPORT_OK = qw{ update_dynamic_config_parameters };
+    our @EXPORT_OK =
+      qw{ update_dynamic_config_parameters update_with_dynamic_config_parameters };
 }
 
 sub update_dynamic_config_parameters {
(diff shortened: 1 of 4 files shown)


lib/MIP/Update/Parameters.pm
```diff
@@ -65,43 +65,84 @@ sub update_dynamic_config_parameters {
...
+            next KEY if ( not defined $active_parameter_href->{$parameter_name}{$key} );
+
+            if (   ref $active_parameter_href->{$parameter_name}{$key} eq q{HASH}
+                or ref $active_parameter_href->{$parameter_name}{$key} eq q{ARRAY} )
+            {
+
+                update_dynamic_config_parameters(
+                    {
+                        active_parameter_href =>
+                          $active_parameter_href->{$parameter_name},
```
henrikstranneheim, line 86: Should you not go down one level here?
```suggestion
                        active_parameter_href =>
                          $active_parameter_href->{$parameter_name}{key},
```

lib/MIP/Update/Parameters.pm
```diff
@@ -65,43 +65,84 @@ sub update_dynamic_config_parameters {
...
+        {
+
+            next ELEMENT if ( not $element );
+
+            if ( ref $element eq q{HASH} or ref $element eq q{ARRAY} ) {
+
+                update_dynamic_config_parameters(
+                    {
+                        active_parameter_href =>
+                          $active_parameter_href->{$parameter_name},
```
henrikstranneheim, line 119: Same here? Otherwise would you not "start over" at the same place in the data structure?

lib/MIP/Update/Parameters.pm
jemten: Isn't that what I'm doing by setting $active_parameter_href to $active_parameter_href->{$parameter_name}? Feels like adding $key also I would move down two steps. I'l take a closer look though. I'm easily confused by these recursive subs 😕

henrikstranneheim: I think you want to go in at that level. You will return once the data is exhausted

henrikstranneheim: But write proper test is the way to test all this and some print statements

jemten:
Was a bit more complex than I first thought. Will rewrite this

lib/MIP/Update/Parameters.pm
```diff
@@ -129,21 +173,11 @@ sub update_dynamic_config_parameters {
             {
 
                 ## Replace dynamic config parameters with actual value that is now set from cmd or config
-                $active_parameter_href->{$parameter_name}[$element_index] =~
+                $active_parameter_href->[$element_index] =~
```
henrikstranneheim, line 176: Will you add the sub call here, which you created above?

jemten: Good idea!

jemten:
I updated the sun to use scalar references. Think I'm done with this for now

PR #1493, opened 2020-04-24
feat(refactor): compression features

henrikstranneheim:
### This PR fixes:

- Refactor how to get and set compression features

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/File_info.pm
+++ b/lib/MIP/File_info.pm
@@ -42,6 +42,7 @@ BEGIN {
       set_sample_file_attribute
       set_select_file_contigs
       parse_file_compression_features
+      parse_files_compression_status
       parse_select_file_contigs
     };
 }
@@ -217,7 +218,7 @@ sub get_sample_file_attribute {
     my $tmpl = {
         attribute => {
             allow => [
-                qw{ is_file_uncompressed
+                qw{ is_file_compressed
                   read_file_command
                   }
             ],
@@ -253,7 +254,8 @@ sub get_sample_file_attribute {
         return %{ $file_info_href->{$sample_id}{$file_name} };
     }
     ## Get attribute
-    my $stored_attribute = $file_info_href->{$sample_id}{$file_name}{$attribute};
+    my $stored_attribute =
+      $file_info_href->{$sample_id}{$file_name}{$attribute};
 
     ## Return requested attribute
     return $stored_attribute;
@@ -261,7 +263,7 @@ sub get_sample_file_attribute {
 
 sub parse_file_compression_features {
 
-## Function : Set file compression features
+## Function : Parse file compression features
 ## Returns  : $attribute{read_file_command}
 ## Arguments: $file_info_href => File info hash {REF}
 ##          : $file_name      => File name
@@ -311,37 +313,8 @@ sub parse_file_compression_features {
     ## Gzipped
     if ($is_gzipped) {
 
-        my $is_files_compressed = get_is_sample_files_compressed(
-            {
-                file_info_href => $file_info_href,
-                sample_id      => $sample_id,
-            }
-        );
-
-        if ( not defined $is_files_compressed ) {
-
-            ## Set is_files_compressed per sample global boolean
-            set_is_sample_files_compressed(
-                {
-                    compression_status => 1,
-                    file_info_href     => $file_info_href,
-                    sample_id          => $sample_id,
-                }
-            );
-
-        }
         $attribute{is_file_compressed} = 1;
         $attribute{read_file_command}  = q{gzip -d -c};
-    }
-    else {
-        ## Set is_files_compressed per sample global boolean
-        set_is_sample_files_compressed(
-            {
-                compression_status => 0,
-                file_info_href     => $file_info_href,
-                sample_id          => $sample_id,
-            }
-        );
     }
 
   ATTRIBUTES:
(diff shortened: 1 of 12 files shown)


lib/MIP/File_info.pm
```diff
@@ -151,6 +156,210 @@ sub check_parameter_metafiles {
...
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    use Data::Diver qw{ Dive };
+
+    if ( defined Dive( $file_info_href, ( q{is_files_compressed}, $sample_id ) ) ) {
+
+        ## Return files compression status
+        return $file_info_href->{is_files_compressed}{$sample_id};
```
jemten, line 195: Shouldn't this be set on file level rather than sample_id level. Is there anything stopping us from having a mix of compressed and uncompressed infiles for one sample_id?

lib/MIP/File_info.pm
```diff
@@ -151,6 +156,210 @@ sub check_parameter_metafiles {
...
+    }
+    return;
+}
+
+sub get_sample_file_attribute {
```
jemten, line 200: Could the sub above be merged into this one?

lib/MIP/File_info.pm
```diff
@@ -151,6 +156,210 @@ sub check_parameter_metafiles {
...
+        read_file_command  => q{cat},
+    );
+
+    ## Check if a file is gzipped.
+    my $is_gzipped = check_gzipped( { file_name => $file_name, } );
+
+    ## Gzipped
+    if ($is_gzipped) {
+
+        my $is_files_compressed = get_is_sample_files_compressed(
```
jemten, line 314: If the file is gzipped as we have checked above then it's compressed. Is this check necessary?

lib/MIP/File_info.pm
```diff
@@ -151,6 +156,210 @@ sub check_parameter_metafiles {
...
+        if ( not defined $is_files_compressed ) {
+
+            ## Set is_files_compressed per sample global boolean
+            set_is_sample_files_compressed(
+                {
+                    compression_status => 1,
+                    file_info_href     => $file_info_href,
+                    sample_id          => $sample_id,
+                }
+            );
```
jemten, line 330: Isn't statement a bit strange? $is_files_compressed should be 1 if the file is compressed. I'm interpreting this as if the file is uncompressed we're setting the compression status to 1 i.e. it is compressed.

lib/MIP/File_info.pm
henrikstranneheim: Actually, there is both.
- Per file is set in the sub set_sample_file_attribute
- All files per samples compression status is set in set_is_sample_files_compressed

Both are used. The overall to avoid going into the gzip_fastq analysis recipe before setting io outfiles and the other should be used (but probably is not yet) in the gzip_fastq recipe. However, it does seem handy to have it at file level also.

lib/MIP/File_info.pm
henrikstranneheim: No, I do not think so they operate at diffeent levels: one per sample and one per sample and file

lib/MIP/File_info.pm
henrikstranneheim: Yes, I think so, since we need to set it at both sample-file level and sample level

lib/MIP/File_info.pm
henrikstranneheim: It is to cater for a mix of compressed and uncompressed files and we want to know if a single one is uncompressed then $is_files_compressed is defined by the else below and we do not want to reset it to one since they are then not all compressed. I think I got the logic correct, but it took some time to wrap my head around it, If you have a more clear solution it would be great because this one does seem to require some thinking.

henrikstranneheim: I guess one thing one probably can do is to parse per sample and file later in another sub looking at the `is_file_compressed`. That might make it easier to write clearer logic

lib/MIP/File_info.pm
jemten: OK, makes sense. So a mix of compressed and uncompressed files for a sample_id would return $file_info_href->{is_files_compressed}{$sample_id} = 0? As you said we should add the file check to the gzip recipe

lib/MIP/File_info.pm
jemten: 👍

lib/MIP/File_info.pm
jemten: I think that something like that might be needed here. I can give it some more thought on Monday. Go ahead and merge if you want to and we can discuss on Monday how this logic can be made easier for the next time we need to wrap our heads around it. I have no doubts that the code works and does what it needs to 😄

henrikstranneheim: @jemten I moved the logic to a separate parsing sub and after all sample files has been processed. I think it became a lot better. Please have a look

lib/MIP/File_info.pm
```diff
@@ -217,7 +218,7 @@ sub get_sample_file_attribute {
     my $tmpl = {
         attribute => {
             allow => [
-                qw{ is_file_uncompressed
+                qw{ is_file_compressed
```
jemten, line 221: 👍 like to switch to check for compressed instead of uncompressed

henrikstranneheim: Are you looking at old code? I thought I did that?

henrikstranneheim: yes, i believe this is fixed cannot find it either in my local copy or centrally

jemten: I don't know. It popped up as a change since last

PR #1546, opened 2020-06-04
Feature/set file direction attributes in sample info

henrikstranneheim:
### This PR adds | fixes:

- See commit msg below

### How to test:

### Expected outcome:
- [x] Installation, unit and integration test suite pass

### Review:
- [x] Code review
- [x] Tests pass


lib/MIP/Sample_info.pm
```diff
@@ -904,14 +900,15 @@ sub set_infile_info {
...
-        }
+        my %direction_two_metric = ( sequence_run_type => q{paired-end}, );
+        set_no_read_direction_file_attributes(
+            {
+                file_name              => $mip_file_format,
+                no_read_direction_href => \%direction_two_metric,
+                sample_id              => $sample_id,
+                sample_info_href       => $sample_info_href,
+            }
+        );
```
jemten, line 911: 👍

lib/MIP/Sample_info.pm
```diff
@@ -925,17 +922,15 @@ sub set_infile_info {
...
-
+    set_read_direction_file_attributes(
+        {
+            direction_file_name => $mip_file_format_with_direction,
+            file_name           => $mip_file_format,
+            read_direction_href => \%both_directions_metric,
+            sample_id           => $sample_id,
+            sample_info_href    => $sample_info_href,
+        }
+    );
```
jemten, line 933: 👍

templates/code/test.t
```diff
@@ -59,6 +59,8 @@ diag(   q{Test SUB_ROUTINE from MODULE_NAME.pm v}
 
 ## Given
 
+## When
+
```
jemten, line 63: I know that it is good practice to use "When" also, but often I find myself with a more ore less static expression for the when part. Something in the line of  "## When executing subroutine". Is it helpful then? You don't use it in the tests linked to this PR

henrikstranneheim: I know. It is usually because the sub is so small and pure i.e. there is no logical conditions in it like if statements. I think it good to have it in the template because it is part of the global style of writing these types of test. As a nudge to use it (if applicable)

jemten: I'm fine with having it in the template :)
But then we don't enforce using it in the tests?

henrikstranneheim: If it is not relevant and adds useful information I think we should skip it. Then it would only clutter the flow

PR #1626, opened 2020-08-19
feat(download): build_fai

jemten:
### This PR adds | fixes:

- builds the fai file during download 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Recipes/Download/Human_reference.pm
+++ b/lib/MIP/Recipes/Download/Human_reference.pm
@@ -8,14 +8,12 @@ use File::Basename qw{ dirname };
 use File::Spec::Functions qw{ catfile };
 use open qw{ :encoding(UTF-8) :std };
 use Params::Check qw{ allow check last_error };
-use strict;
 use utf8;
 use warnings;
 use warnings qw{ FATAL utf8 };
 
 ## CPANM
 use autodie qw{ :all };
-use Readonly;
 
 ## MIPs lib/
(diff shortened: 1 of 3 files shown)


lib/MIP/Recipes/Download/Human_reference.pm
```diff
@@ -186,6 +185,19 @@ sub download_human_reference {
         }
     );
 
+    my $outfile_path =
+      catfile( $active_parameter_href->{reference_dir}, $reference_href->{outfile} );
+    my $outfile_no_gz =
+      parse_file_suffix( { file_name => $outfile_path, file_suffix => q{.gz}, } )
+      // $outfile_path;
```
henrikstranneheim, line 192: What is `//`?
Should it not be `||=`?

henrikstranneheim (changes requested):
👍

lib/MIP/Recipes/Download/Human_reference.pm
dnil: If there ever was a confusion when moving between languages: comment codes from other languages ought to have some kind of elevated status.. 😄

dnil: https://perldoc.perl.org/perlop.html#Logical-Defined-Or

jemten: Hehe, always fun to dig out some obscure operator from the perl drawer. The `//`check if the left hand is defined otherwise it returns the right hand side.

PR #1695, opened 2020-10-16
Feature/only deepvar

ramprasadn:
### This PR adds | fixes:

- improves glnexus test coverage
- Update gatk combinevariants recipe to work with glnexus output  

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


lib/MIP/Recipes/Analysis/Gatk_combinevariantcallsets.pm
```diff
@@ -232,12 +232,21 @@ sub analysis_gatk_combinevariantcallsets {
...
         my $infile_path_prefix = $sample_io{$stream}{file_path_prefix};
-        my $infile_suffix      = $sample_io{$stream}{file_suffix};
+        my $infile_suffix      = $sample_io{$stream}{file_suffixes}[0];
         my $infile_path        = $infile_path_prefix . $infile_suffix;
 
         ## Only use first part of name
-        my ($variant_caller_prio_tag) = split /_/sxm, $variant_caller;
+        my $variant_caller_prio_tag;
 
+        if ( $variant_caller eq q{gatk_variantrecalibration} ) {
```
henrikstranneheim, line 241: It is often better and more extendible to avoid if-else trees and try other techniques like a hash map, for instance:
```
%variant_caller_tag_map = (gatk_variantrecalibration => q{haplotypecaller},
glnexus_merge => q{deepvariant},
bcftools_mpileup => q{mpileup},);
```
and then 
```
$variant_caller_prio_tag = $variant_caller_tag_map{$variant_caller};
```

lib/MIP/Recipes/Analysis/Gatk_combinevariantcallsets.pm
```diff
@@ -232,11 +232,19 @@ sub analysis_gatk_combinevariantcallsets {
...
-        my ($variant_caller_prio_tag) = split /_/sxm, $variant_caller;
+        my $variant_caller_prio_tag;
+
+        my %variant_caller_tag_map = (
+            gatk_variantrecalibration => q{haplotypecaller},
+            glnexus_merge             => q{deepvariant},
+            bcftools_mpileup          => q{mpileup},
+        );
+
+        $variant_caller_prio_tag = $variant_caller_tag_map{$variant_caller};
```
jemten, line 247:
```suggestion
        my $variant_caller_prio_tag = $variant_caller_tag_map{$variant_caller};
```

lib/MIP/Recipes/Analysis/Gatk_combinevariantcallsets.pm
```diff
@@ -232,11 +232,19 @@ sub analysis_gatk_combinevariantcallsets {
...
         );
         my $infile_path_prefix = $sample_io{$stream}{file_path_prefix};
-        my $infile_suffix      = $sample_io{$stream}{file_suffix};
+        my $infile_suffix      = $sample_io{$stream}{file_suffixes}[0];
         my $infile_path        = $infile_path_prefix . $infile_suffix;
 
         ## Only use first part of name
-        my ($variant_caller_prio_tag) = split /_/sxm, $variant_caller;
+        my $variant_caller_prio_tag;
+
```
jemten, line 240:
```suggestion
```

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -131,16 +131,13 @@ sub analysis_glnexus {
 
     ## Unpack parameters
     my $consensus_analysis_type;
```
jemten, line 133:
```suggestion
```

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -131,16 +131,13 @@ sub analysis_glnexus {
...
-        $consensus_analysis_type = q{WES};
-    }
-    elsif ( uc $parameter_href->{cache}{consensus_analysis_type} eq q{MIXED} ) {
-        $consensus_analysis_type = q{WGS};
-    }
-    else {
-        $consensus_analysis_type =
-          uc( $parameter_href->{cache}{consensus_analysis_type} );
-    }
+
```
jemten, line 134:
```suggestion
```

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -131,16 +131,13 @@ sub analysis_glnexus {
...
+    my %consensus_analysis_type_map =
+      ( MIXED => q{WGS}, PANEL => q{WES}, WGS => q{WGS}, WES => q{WES} );
+
+    $consensus_analysis_type =
```
jemten, line 138:
```suggestion
    my $consensus_analysis_type =
```

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -225,24 +224,19 @@ sub analysis_glnexus {
             dir        => catdir( $active_parameter_href->{temp_directory}, q{glnexus} ),
```
jemten, line 224: No threads option?

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -225,24 +224,19 @@ sub analysis_glnexus {
...
-    print {$filehandle} $PIPE . $SPACE;
-
-    bcftools_view(
-        {
-            filehandle  => $filehandle,
-            infile_path => $DASH,
-        }
-    );
-    print {$filehandle} $PIPE . $SPACE;
+    print {$filehandle} $NEWLINE . $NEWLINE . q{## View} . $NEWLINE;
```
jemten, line 230:
```suggestion
    say {$filehandle} $NEWLINE;
     
    say {$filehandle} q{## View};
```

lib/MIP/Recipes/Analysis/Glnexus.pm
```diff
@@ -17,15 +17,15 @@ use autodie qw{ :all };
 use Readonly;
 
 ## MIPs lib/
-use MIP::Constants qw{ $DASH $LOG_NAME $NEWLINE $PIPE $SPACE $UNDERSCORE };
+use MIP::Constants qw{ $DASH $LOG_NAME $NEWLINE $PIPE $SEMICOLON $SPACE $UNDERSCORE };
```
jemten, line 20:
```suggestion
use MIP::Constants qw{ $DASH $LOG_NAME $NEWLINE $PIPE $SPACE $UNDERSCORE };
```

jemten: Doesn't seem to be in use

jemten (commented):
Great work Ram! Nearly there

lib/MIP/Program/Gatk.pm
```diff
@@ -26,7 +26,7 @@ BEGIN {
     use base qw{ Exporter };
 
     # Set the version for version checking
-    our $VERSION = 1.21;
+    our $VERSION = 1.22;
```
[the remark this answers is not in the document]
ramprasadn: Sorry about this one! I kept going back and forth on making changes in this file and the version number slipped through somehow.

definitions/rd_dna_parameters.yaml
```diff
@@ -160,7 +160,7 @@ recipe_core_number:
     gatk_variantevalall: 1
     gatk_variantevalexome: 1
     gatk_variantrecalibration: 1
-    glnexus_merge: 36
+    glnexus_merge: 1
```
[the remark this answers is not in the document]
ramprasadn: Glnexus is multithreaded indeed, but by default it will make use of whatever resource is available at its disposal. 
I realized that this cpu number was being used to create the ntasks variable for the slurm script. But at any given time we are only running one task at a time (First - glnexus, then view, and then index). So it didn't make sense to have 36 there. Having said that once the task starts, it will create 36 threads for the glnexus command because that's the default value set in the code. And these 36 threads will figure out a way to run using the resources efficiently. 

Under current configuration glnexus step only took about 14 mins to run and it also seemed like the job didn't have to wait in queue for long. (It is entirely possible that I am missing something with regards to ntasks variable here, so please correct me if I am wrong)

definitions/rd_dna_parameters.yaml
```diff
@@ -217,9 +217,11 @@ recipe_memory:
     expansionhunter: 2
     variant_annotation: 3
     gatk_baserecalibration: 10
+    gatk_combinevariantcallsets: 25
     gatk_gathervcfs: 7
     gatk_genotypegvcfs: 12
     gatk_variantrecalibration: 30
+    glnexus_merge: 150
```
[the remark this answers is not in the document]
ramprasadn: While running on the login node, I noticed that the allocated memory spiking over 100Gb for about a minute or two. So to play it safe, I have specified 150GB here.

lib/MIP/Recipes/Analysis/Glnexus.pm
ramprasadn: Glnexus program recipe has 36 threads as the default value, but I thought it would be redundant to set the value in the analysis recipe too. Con is that when there is a need to change the number of threads, one would have to do it in the program recipe. But I am also not aware of a parameter in the config where one can set the number of threads. I am not sure whats the best way forward here. What do you guys think?

lib/MIP/Program/Glnexus.pm
```diff
@@ -86,13 +88,20 @@ sub glnexus_merge {
             store       => \$stdoutfile_path,
             strict_type => 1,
         },
+        threads => {
+            default     => 36,
```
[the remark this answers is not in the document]
jemten: Let's remove the default and put the option under an if statement. The default of glnexus is to use all available threads and memory, which seems like the desired way to run the program to me. [Glnexus performance](https://github.com/dnanexus-rnd/GLnexus/wiki/Performance)

definitions/rd_dna_parameters.yaml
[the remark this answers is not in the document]
jemten: That's true. Didn't think about that. So even if we allocate 36 cores it will only run on one core? Maybe we should introduce the sbatch header '--cpus-per-task' in a later PR. However, since we are running multiple programs in a pipe wouldn't that speed up the process? We are allocating so much memory at the moment so the cores are basically free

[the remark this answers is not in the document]
ramprasadn: Looking at the time taken to run glnexus on justhusky, my impression is that number of threads matter more than the number of cores. I think using 1 core also adds another benefit as in glnexus is likely to find a free core to run rather soon than to wait in the queue for 10 or 36 cores to be free. 
Also, those other programs only take about a minute to run, as glnexus output isn't that big.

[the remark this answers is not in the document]
jemten: The 150G of memory will be the bottleneck here. For many programs we don't specify the amount of memory and I don't think that balsamic specify it at all, or at least they din't use to. By default each core gets 5G so this is equivalent of allocating 30 cores. But sure, if there are no performance benefits to more cores this is the most correct.

jemten (approved):
Looks great!

PR #1713, opened 2020-11-10
feat(conda_perl): Using conda version 4.9.1 this will install all MIP…

henrikstranneheim:
… perl dependencies without errors using only cpanm

### This PR adds:

- See above. Placing this file here for future reference. Showing that using an updated version of conda the installation actually runs quite smooth.

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/templates/mip_install_perl_conda-v4.9.1.sh
+++ b/templates/mip_install_perl_conda-v4.9.1.sh
@@ -1,0 +1,87 @@
+#! /usr/bin/env/bash
+set -e
+
+usage="$(basename "$0") [-e <environment_name> -p <conda_path> -x <existing_conda_env> -d <install dev tools>]
+
+where:
+    -h  Show this help text
+    -e  Conda environment name [string]
+    -p  Conda path [string]
+    -x  Install in already existing conda environment [flag]
+    -d  Install developer tools"
+
+unset OPTARG
+unset OPTIND
+
+CONDA_PATH="$HOME/miniconda3"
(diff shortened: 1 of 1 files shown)


templates/mip_install_perl_conda-v4.9.1.sh
```diff
@@ -0,0 +1,87 @@
...
+if [ "$DEV" = true ]; then
+
+    ## Install dev modules from cpan
+    cpanm Perl::Tidy@20200110 Perl::Critic@1.138 Data::Printer@0.40
+
+    ## Install yamllint
+    conda install --name "$ENV_NAME" --yes -c conda-forge yamllint=1.20.0
+fi
+
+#conda deactivate
```
jemten, line 87: remove?

jemten (approved):
Do we need to update conda on hasta for this to work?

henrikstranneheim:
> Do we need to update conda on hasta for this to work?

Yes, this is more for bookkeeping that this is all that should be required with well updated and rather clean conda.

templates/mip_install_perl_conda-v4.9.1.sh
henrikstranneheim, line 87:
```suggestion
conda deactivate
```

henrikstranneheim: Removed comment symbol

PR #1742, opened 2020-12-07
Feature/carton

henrikstranneheim:
### This PR adds | fixes:

- Adds a cpanfile snapshot created by using Carton. This creates a map aof all MIP's perl dependecy modules
- Use carton in MIPs dockerfile to always use exact version of perl dependencies when installing modules and running MIP in deployment
- Adds instructions in the CONTRIBUTING on Carton adn Carmel

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ mip_log/
 **/.idea
 _Deparsed_XSubs.pm
 .gitignore
+local/
+local/
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,9 @@ mip_log/
 **/.idea
 _Deparsed_XSubs.pm
 .gitignore
+
+## Carton
 local/
-local/
+
+## Carmel
+.carmel/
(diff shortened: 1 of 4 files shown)


jemten (approved):
Fantastic!
Maybe explain carmel a bit more. Are they interchangeable? 
Shouldn't we update the mip_install_perl.sh file with this carton procedure instead?

ramprasadn (approved):

henrikstranneheim:
> Maybe explain carmel a bit more. Are they interchangeable?

Somewhat but they differ in where they place the libs and how they are choosen for the application. Carmel is also so far experimental. I think I will leave it as it is for now.

> Shouldn't we update the mip_install_perl.sh file with this carton procedure instead?

I would like to wait until we use it like that in production. Right now it is only used in the dockerfile to make a locked deployment version of MIP and provide a wrapper to execute MIP via carton instead of directly.

PR #1749, opened 2020-12-10
Feature/pre commit

jemten:
### This PR adds | fixes:

- Replaces mip-check with pre-commit

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim (approved):
Beautiful

mip_install_perl.sh
```diff
@@ -77,13 +77,16 @@ conda activate "$ENV_NAME"
 
 cpanm --installdeps .
 
-if [ "$DEV" = true ]; then 
-    
+if [ "$DEV" = true ]; then
+
     ## Install dev modules from cpan
-    cpanm Perl::Tidy@20200110 Perl::Critic@1.138 Data::Printer@0.40
+    cpanm Perl::Critic@1.138 Data::Printer@0.40
```
henrikstranneheim, line 83: What happened to perl tidy here?

jemten: pre-commit will install it's own version of perltidy as well as yamllint. The Perlcritic repo don't have a .pre-commit-hooks.yaml file and thus we have to use the one in the conda env. You might still want perltidy and yamllint so that you can add it to your IDE but it's not needed for pre-commit.

jemten: I could add it back though. I will probably have it in my dev environment

ramprasadn (approved):
Nice!! Looking good 👍

mip_install_perl.sh
henrikstranneheim: Sounds good. And maybe add a sentance on pre-commit will use its own install in and the IDE the other one in contributing doc

PR #1752, opened 2020-12-14
Release/9.1.1

jemten:
### This PR adds | fixes:

- Fixes MIP gender estimation

SINGULARITY_BIND isn't set when MIP parses the fastqs for gender, causing the command to not find the reference. This leads to an error and MIP reports this as a zero reads found mapping to Y. This in turn leads to MIP classifying all wgs samples with unknown gender as female. 
image

This PR sets SINGULARITY_BIND for the bwa command so that the command works as expected. 

### How to test:

- Run the case hotalpaca in dry run mode and see that MIP classifies the sample as male

- Check that SINGULARITY_BIND has it's original value after MIP execution

### Expected outcome:
- [x] MIP classifies the sample as male
- [x] SINGULARITY_BIND has its original value 

### Review:
- [ ] Code review
- [ ] Tests pass


jemten:
Check SINGULARITY_BIND and start ✔️ 
image
SINGULARITY_BIND at its original value ✔️ 
image

henrikstranneheim (approved):
So that was the culprit! Well done!

PR #1775, opened 2021-01-10
Get recipe resources

henrikstranneheim:
### This PR adds | fixes:

- Move subs: get_recipe_resources and get_recipe_attributes and refactor them and their corresponding tests
- Create new sub to parse recipe prerequisites: parse_recipe_prerequisites in Recipe.pm
- Remove get_recipe_attributes and get_recipe_resources subs from recipes and use parse_recipe_prerequisites instead
- Use hash from parse_recipe_prerequisites to access recipe data in recipes

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


lib/MIP/Recipe.pm
```diff
@@ -0,0 +1,96 @@
...
+        active_parameter_href => {
+            default     => {},
+            defined     => 1,
+            required    => 1,
+            store       => \$active_parameter_href,
+            strict_type => 1,
+        },
+        parameter_href => {
+            default     => {},
+            defined     => 1,
```
jemten, line 50: Is defined needed here? Looks like you can use it without $parameter_href or at least an empty one

lib/MIP/Recipe.pm
```diff
@@ -0,0 +1,96 @@
...
+            store       => \$parameter_href,
+            strict_type => 1,
```
jemten, line 52: order ab

lib/MIP/Recipe.pm
```diff
@@ -0,0 +1,96 @@
...
+        },
+        recipe_name => {
+            defined     => 1,
+            required    => 1,
+            strict_type => 1,
+            store       => \$recipe_name,
```
jemten, line 58: order ab

jemten (approved):
Very nice addition. I like how it simplifies the recipes 😄

lib/MIP/Active_parameter.pm
```diff
@@ -1055,6 +1056,87 @@ sub get_package_env_cmds {
...
+    my %recipe_resource = (
+        core_number  => $core_number,
+        gpu_number   => $active_parameter_href->{recipe_gpu_number}{$recipe_name},
+        load_env_ref => \@environment_cmds,
+        memory       => $memory,
+        mode         => $active_parameter_href->{$recipe_name},
+        time         => $active_parameter_href->{recipe_time}{$recipe_name},
+    );
+
+    return $recipe_resource{$resource} if ( defined $resource && $resource );
```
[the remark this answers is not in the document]
henrikstranneheim: You are right and also `undef` is not allowed as a value so the process would not even get here. I will remove it

lib/MIP/Recipe.pm
henrikstranneheim: It is not required so it is ok to not supply it. However, if you do supply it it has to be defined. I think it is ok to leave it as it is.

lib/MIP/Recipe.pm
henrikstranneheim: fixed

PR #1805, opened 2021-02-05
feat(refactor): Move sub get_number_of_male_reads

henrikstranneheim:
### This PR adds | fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Parse/Gender.pm
+++ b/lib/MIP/Parse/Gender.pm
@@ -5,7 +5,6 @@ use Carp;
 use charnames qw{ :full :short };
 use English qw{ -no_match_vars };
 use File::Spec::Functions qw{ catfile };
-use FindBin qw{ $Bin };
 use open qw{ :encoding(UTF-8) :std };
 use Params::Check qw{ allow check last_error };
 use utf8;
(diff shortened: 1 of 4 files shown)


lib/MIP/Parse/Gender.pm
```diff
@@ -18,87 +17,25 @@ use Readonly;
...
-            commands_ref => $cmds_ref,
-            process_type => q{ipc_cmd_run},
-        }
-    );
-
-    my $y_read_count = $process_return{stdouts_ref}->[0];
-
-    ## Clean-up
-    close $filehandle;
-    remove_tree($bash_temp_file);
```
jemten, line 97: Just curious. Why do we delete the file after we're done? Also, we could add `tee` to the command in order to save the output of the command to a file as well. But again the better option here is probably to rewrite this module a bit in order to log some more values and what command has been run. Another PR :)

jemten (approved):
👍

lib/MIP/Parse/Gender.pm
henrikstranneheim: Yeah, I also wondered that. It seems wrong as one removes the record of what was done making it hard to reproduce and inspect

henrikstranneheim:
@jemten I made some changes - could you give it another look?

jemten (approved):
Nice! Now we have a track record of the command that was run

lib/MIP/Parse/Gender.pm
henrikstranneheim: Fixed, except tee suggestion

PR #1808, opened 2021-02-09
Feature/gres update

ramprasadn:
### This PR adds | fixes:

- Updates gpu headers from partition to gres

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/lib/MIP/Program/Slurm.pm
+++ b/lib/MIP/Program/Slurm.pm
@@ -179,8 +179,6 @@ sub slurm_build_sbatch_header {
 
     if ($gpu_number) {
 
-        #        push @commands, q{--partition=} . q{gpu};
-        #        push @commands, q{--gpus=} . $gpu_number;
         push @commands, q{--gres=gpu};
     }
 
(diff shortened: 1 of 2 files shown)


lib/MIP/Program/Slurm.pm
```diff
@@ -178,8 +178,8 @@ sub slurm_build_sbatch_header {
     }
 
     if ($gpu_number) {
-        push @commands, q{--partition=} . q{gpu};
-        push @commands, q{--gpus=} . $gpu_number;
+
+        push @commands, q{--gres=gpu};
```
jemten, line 182: Is there a way of setting gpu number with this way of allocating gpu nodes?

ramprasadn: Yes, that's possible! Even though deepvariant can't use more than one GPU, it would be a nice to future proof this bit. Will make the change in the next commit. :D

jemten (approved):
👍

PR #1813, opened 2021-02-17
Implement memory limit for glnexus

ramprasadn:
### This PR adds | fixes:

- Glnexus crashes due to memory conflicts

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


t/glnexus_merge.t
```diff
@@ -60,6 +59,10 @@ my %base_argument = (
         inputs_ref      => [ catfile(qw{ dir infile1.vcf }) ],
         expected_output => catfile(qw{ dir infile1.vcf }),
     },
+    memory => {
+        input           => q{10},
```
jemten, line 63: Define a readonly constant for numbers other than 0, 1, 2. Like this:
```Perl
Readonly my $MEMORY => 10;
```
And use that instead of `q{10}`
Might seem a bit excessive but we're trying to adhere to it elsewhere

t/glnexus_merge.t
```diff
@@ -100,6 +103,10 @@ my %specific_argument = (
...
+        expected_output => q{--mem-gbytes } . q{10},
```
jemten, line 108: same here

jemten (approved):
Looks good Ram 👍

jemten:
Does glnexus work with these settings? If that's the case, I think we should merge this PR

henrikstranneheim:
@ramprasadn bump

PR #1830, opened 2021-03-04
Feature/serial fusion calling

jemten:
### This PR adds | fixes:
- Run the fusion caller in serial mode instead of parallel. 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/definitions/rd_rna_initiation_map.yaml
+++ b/definitions/rd_rna_initiation_map.yaml
@@ -10,9 +10,8 @@ CHAIN_ALL:
     - salmon_quant
     - blobfish
   - CHAIN_FUSION:
-    - PARALLEL:
-      - star_fusion
-      - arriba_ar
+    - star_fusion
+    - arriba_ar
     - megafusion_ar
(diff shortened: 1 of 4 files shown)


definitions/rd_rna_initiation_map.yaml
```diff
@@ -10,9 +10,8 @@ CHAIN_ALL:
     - salmon_quant
     - blobfish
   - CHAIN_FUSION:
-    - PARALLEL:
-      - star_fusion
-      - arriba_ar
+    - star_fusion
```
henrikstranneheim, line 13: I guess this makes it a lot easier without prolonging the run time that much 👍

henrikstranneheim (approved):
Nice!

definitions/rd_rna_initiation_map.yaml
jemten: Yeah, I think this is OK in the short term at least.

PR #1835, opened 2021-03-08
Feature/reference check vt

ramprasadn:
### This PR adds | fixes:

- bcftools norm for normalizing reference vcfs

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


lib/MIP/Check/Reference.pm
```diff
@@ -61,76 +59,73 @@ sub check_if_processed_by_vt {
...
+    use MIP::File::Format::Vcf qw{ get_bcftools_norm_command_from_vcf_header };
 
     ## Retrieve logger object
     my $log = Log::Log4perl->get_logger($LOG_NAME);
 
-    my %vt_regexp;
-    $vt_regexp{vt_decompose}{vcf_key} = q{OLD_MULTIALLELIC};
-    $vt_regexp{vt_normalize}{vcf_key} = q{OLD_VARIANT};
+    my %bcftools_regexp;
+    $bcftools_regexp{bcftools_normalize}{vcf_key} = q{bcftools_norm};
```
henrikstranneheim, line 68: This could be in one line using something like:
```suggestion
    my %bcftools_regexp = (bcftools_normalize => {vcf_key => q{bcftools_norm},},);
```
However, I if you do not see us adding more regexpes in the near future -  I suggest you remove the foor loop below and just call the sub using with header_id => q{bcftools_norm}

lib/MIP/Check/Reference.pm
```diff
@@ -61,76 +59,73 @@ sub check_if_processed_by_vt {
...
 
     my @to_process_references;
-    ## Downloaded and vt later (for downloadable references otherwise
+    ## Downloaded and check bcftools header later (for downloadable references otherwise
     ## file existens error is thrown downstream)
     if ( not -e $reference_file_path ) {
 
         ## Do nothing since there is not ref file to check
         return;
     }
```
henrikstranneheim, line 77:
```suggestion
    return if ( not -e $reference_file_path );
```

lib/MIP/Check/Reference.pm
```diff
@@ -61,76 +59,73 @@ sub check_if_processed_by_vt {
...
-    foreach my $vt_parameter_name ( keys %vt_regexp ) {
+  BCFTOOLS_PARAMETER_NAME:
+    foreach my $bcftools_parameter_name ( keys %bcftools_regexp ) {
 
-        my $header_id_line = get_vcf_header_line_by_id(
+        my $header_id_line = get_bcftools_norm_command_from_vcf_header(
             {
                 bcftools_binary_path => $bcftools_binary_path,
-                header_id            => $vt_regexp{$vt_parameter_name}{vcf_key},
+                header_id            => $bcftools_regexp{$bcftools_parameter_name}{vcf_key},
```
henrikstranneheim, line 85: See comment above

lib/MIP/Check/Reference.pm
```diff
@@ -178,11 +173,10 @@ sub check_references_for_vt {
     );
 
   PARAMETER_NAME:
-    foreach my $parameter_name ( @{$vt_references_ref} ) {
+    foreach my $parameter_name ( @{$bcftools_references_ref} ) {
 
       ASSOCIATED_RECIPE:
-        foreach my $associated_recipe (
-            @{ $parameter_href->{$parameter_name}{associated_recipe} } )
+        foreach my $associated_recipe ( @{ $parameter_href->{$parameter_name}{associated_recipe} } )
```
henrikstranneheim, line 179: This one knows to much about the parameter hash. Use sub `get_parameter_attribute`to get the associated_recipes array and then loop around this local array.

lib/MIP/Check/Reference.pm
```diff
@@ -224,14 +218,12 @@ sub check_references_for_vt {
                 ## ARRAY reference
 
               ANNOTION_FILE:
-                foreach
-                  my $annotation_file ( @{ $active_parameter_href->{$parameter_name} } )
-                {
+                foreach my $annotation_file ( @{ $active_parameter_href->{$parameter_name} } ) {
```
henrikstranneheim, line 221: One level knowledge is ok 👍

lib/MIP/File/Format/Vcf.pm
```diff
@@ -37,6 +38,73 @@ BEGIN {
...
+        },
+    };
+
+    check( $tmpl, $arg_href, 1 ) or croak q{Could not parse arguments!};
+
+    use MIP::Environment::Child_process qw{ child_process };
+    use MIP::Language::Perl qw{ perl_nae_oneliners };
+
+    my @check_header_cmds;
+
```
henrikstranneheim, line 83: I think `$bcftools_binary_path`can be initated locally in the sub here with
```
 my $bcftools_binary_path = get_executable_base_command( { base_command => q{bcftools}, } );
```
instead of being intilized in `check_references_for_bcftools` in `lib/MIP/Check/Reference.pm`. I think it is only passed around by the subs `check_if_processed_by_bcftools` and `_parse_vcfanno_toml_path` leading to this one. Or what do you think?

t/check_if_processed_by_bcftools.t
```diff
@@ -50,44 +50,43 @@ diag(   q{Test check_if_processed_by_vt from Reference.pm}
...
 
-## Given a reference file path, when needing vt processing
-my $reference_file_path_no_vt =
+## Given a reference file path, when needing bcftools processing
+my $reference_file_path_no_processing =
   catfile( $Bin, qw{ data references grch37_all_wgs_-phase3_v5b.2013-05-02-.vcf.gz } );
 
-## Check if vt has processed references using regexp
-my @checked_references = check_if_processed_by_vt(
+## Check if bcftools has processed references using regexp
```
henrikstranneheim, line 66:
```suggestion
## When checking if bcftools has processed references using regexp
```

t/check_if_processed_by_bcftools.t
```diff
@@ -50,44 +50,43 @@ diag(   q{Test check_if_processed_by_vt from Reference.pm}
...
 
-## Given a reference file path, when not needing vt processing
-my $reference_file_path_vt =
+## Given a reference file path, when not needing bcftools processing
+my $reference_file_path_bcftools =
   catfile( $Bin, qw{ data references grch37_gnomad.genomes_-r2.0.1-.vcf.gz } );
 
-## Check if vt has processed references using regexp
-@checked_references = check_if_processed_by_vt(
+## Check if bcftools has processed references using regexp
```
henrikstranneheim, line 81:
```suggestion
## When checking if bcftools has processed references using regexp
```

t/check_references_for_bcftools.t
```diff
@@ -56,25 +56,20 @@ my %active_parameter_test = (
     gatk_baserecalibration_known_sites => [
```
henrikstranneheim, line 56: Could you add Given-When-Then to these tests. They seem to be missing

t/data/references/qc_regexp_-v1.23-.yaml
```diff
@@ -129,6 +129,4 @@ vcf2cytosure_version:
   version: "perl -nae 'if($_=~/cytosure\\s+(\\d+[.]\\d+[.]\\d+)/xsm) { print $1;last; }' "
 vcfparser:
   version: "perl -nae 'if($_=~/##Software=<ID=vcfParser.pl,Version=(\\d+.\\d+.\\d+)/) {print $1;last;}' "
-vt:
```
henrikstranneheim, line 132: I do not think we should remove this. Rather delete the whole file if it is not used by the tests. And if used by the test see if we can not switch to a more modern version of this file. Latest seem to be version 1.27

t/mip_analysis.test
```diff
@@ -422,13 +422,12 @@ sub read_infile_vcf {
...
+            ## Test bcftools
+            _test_bcftools_in_vcf_header(
                 {
-                    vcf_header_href   => \%vcf_header,
-                    vt_mode           => $active_parameter_href->{vt_ar},
-                    vt_decompose_mode => $active_parameter_href->{vt_decompose},
-                    vt_normalize_mode => $active_parameter_href->{vt_normalize},
+                    vcf_header_href         => \%vcf_header,
+                    bcftools_mode           => $active_parameter_href->{bcftools_norm},
+                    bcftools_normalize_mode => $active_parameter_href->{bcftools_normalize},
```
henrikstranneheim, line 430: Order AB

t/mip_analysis.test
```diff
@@ -394,7 +394,7 @@ sub read_infile_vcf {
                 }
             );
 
-            if (/ INFO\=\<ID\=([^,]+) /sxm) {
+            if (/ (bcftools_norm) /sxm) {
```
henrikstranneheim, line 397: This switches the regexp a lot from a general what comes after the ID to just looking for the bcftools norm. Will this not brake other tests in mip_analysis_tests?

t/mip_analysis.test
```diff
@@ -1176,74 +1175,66 @@ sub _test_sv_combinevariantcallsets_in_vcf_header {
     ## sv_combinevariantcallset has been used
     return if ( not $sv_combinevariantcallsets_mode );
 
-    return if ( not $sv_vt_decompose_mode );
+    return if ( not $sv_decompose_mode );
 
-    ok( exists $vcf_header_href->{INFO}{OLD_MULTIALLELIC}, q{VTDecompose key: OLD_MULTIALLELIC} );
+    ok(
+        exists $vcf_header_href->{INFO}{OLD_MULTIALLELIC},
```
henrikstranneheim, line 1181: Seems incorrect

t/mip_analysis.test
henrikstranneheim: I think you should look at the output of data when running this test from analysisrunstatus and make sure that nothing is broken.

henrikstranneheim (commented):
Great work RAM! I have some minor comments and improvements as well as one concern. But generally very well done!

lib/MIP/Check/Reference.pm
jemten: Just a suggestion, but the vcf_key q{bcftools_norm} could be a part of the oneliner instead of a parameter. Alternatively, we could rename the perl oneliner to something more general, e.g check `vcf_header_for_key`.

I think we should remove the loop now if we ever only loop over it once. It is easy enough to add if we decide to add more bcftools expressions

lib/MIP/Language/Perl.pm
```diff
@@ -351,6 +352,9 @@ sub perl_nae_oneliners {
         add_binary => {
             binary => $oneliner_parameter,
         },
+        bcftools_norm_check => {
+            id => $oneliner_parameter,
+        },
```
jemten, line 357: see comment above

lib/MIP/Recipes/Analysis/Bcftools_core.pm
```diff
@@ -0,0 +1,390 @@
...
+
+    ## Write stderr for xargs process
+    if ( defined $xargs_file_path_prefix && defined $contig ) {
+
+        ## Redirect xargs output to program specific stderr file
+        $stderrfile_path = $xargs_file_path_prefix . $DOT . $contig . $DOT . q{stderr.txt};
+    }
+
+    ## Initate processing
+    if ( not $instream ) {
```
jemten, line 239: Are we still using `$instream` somewhere or was this when we had a rio recipe? Maybe you know @henrikstranneheim

lib/MIP/Recipes/Analysis/Bcftools_core.pm
```diff
@@ -0,0 +1,390 @@
...
+        bcftools_view(
+            {
+                filehandle  => $filehandle,
+                infile_path => $infile_path,
+                output_type => $bcftools_output_type,
+            }
+        );
+    }
+    ## Replace #FORMAT field prior to smart decomposition (variant vcfs)
+    if ($gnu_sed) {
```
jemten, line 250: Also, this one?

lib/MIP/Recipes/Analysis/Bcftools_core.pm
```diff
@@ -0,0 +1,390 @@
...
+        print {$filehandle} $PIPE . $SPACE;
+
+        gnu_sed(
+            {
+                filehandle => $filehandle,
+                script     => q{'s/ID=AD,Number=./ID=AD,Number=R/'},
+            }
+        );
+    }
+    if ($normalize) {
```
jemten, line 262: Isn't this what we always do?

lib/MIP/Recipes/Analysis/Bcftools_core.pm
```diff
@@ -0,0 +1,390 @@
...
+
+        ## Pipe
+        print {$filehandle} $PIPE . $SPACE;
+
+        bcftools_norm(
+            {
+                filehandle        => $filehandle,
+                infile_path       => q{-},
+                remove_duplicates => 1,
+            }
```
jemten, line 283: Can these operations be done in the same `bcftools norm` command or do we need to do them in a serie?

jemten: Also seems like we're missing the ` --fasta-ref` option. Looks like Bcftools won't normalize without it  http://samtools.github.io/bcftools/bcftools.html#norm

lib/MIP/Recipes/Analysis/Bcftools_core.pm
jemten: Or maybe it's good to keep them if we want to use this recipe again?

jemten (commented):
I wrote down some more comments. This recipe is a bit more involved than I first thought. Some are questions for @henrikstranneheim 😄  
Seems like it used to do more when we had the rio mode i MIP but maybe we can simplify it a bit now.

lib/MIP/Recipes/Analysis/Bcftools_core.pm
henrikstranneheim: I think it can be remove together with the `$instream` arg. I do not see that it is used anywhere anymore. It was probably a remnant of the rio recipes before

lib/MIP/Recipes/Analysis/Bcftools_core.pm
henrikstranneheim: This was to make sure that AD was correct in the header before decomposing. I think most callers today will output AD correctly so it might be safe to remove.

lib/MIP/Recipes/Analysis/Bcftools_core.pm
henrikstranneheim: Yes, I also think we can remove the if condition and the arg. This recipe was most likely to generic to begin with and the scope should be reduced.

lib/MIP/Check/Reference.pm
ramprasadn: I have made some changes in the upcoming commit.. please take a look at it and let me know what you guys think.

t/data/references/qc_regexp_-v1.23-.yaml
ramprasadn: Fixed 👍🏻

t/mip_analysis.test
ramprasadn: You are correct.. Now, I have added a separate regex for bcftools norm and included the original one to parse out INFO strings

t/mip_analysis.test
ramprasadn: Fixed!

lib/MIP/Recipes/Analysis/Bcftools_core.pm
ramprasadn: The script worked without it, so I got rid of that bit.

lib/MIP/Recipes/Analysis/Bcftools_core.pm
ramprasadn: Updated this and the one below too

lib/MIP/Recipes/Analysis/Bcftools_core.pm
ramprasadn: Fixed!

lib/MIP/Recipes/Analysis/Bcftools_core.pm
ramprasadn: Good catch! Bcftools barfed when I asked it to remove duplicates and decompose multiallelics in the same command. So I had to split the task over two bcftools commands.

henrikstranneheim (approved):
Well done! There are some github action annotations that need attention as well before merging.

lib/MIP/Check/Reference.pm
jemten: I don't see much changes here. Maybe I'm missing something. 

I still think that it is a bit more complex than what it needs to be as we're not planning to add more bcftools normalization checks (right?). At the moment we only have one check and one vcf key to look for. We could remove the loop and move the key to the oneliner sub. Also no need to run uniq later since we're only using one regexp check.

What you have written will work though so my comments are just that, comments. Feel free to ignore 😄

jemten: Sorry I looked at an old commit 🤦

jemten (approved):

PR #1853, opened 2021-03-25
Feature/bcftools norm fix reference

ramprasadn:
### This PR adds | fixes:

- Skips bcftools normalization check for genbank haplogroups file
Since the reference file in github is already normalized with bcftools, this check was redundant. Hence skipped. 

- Adds an option to fix errors in references
In positions where there is a discrepancy between the vcf and reference fasta, check reference option will try to fix it by modifying the entry in vcf file.

For, example: In the image below, vcf file indicates that the reference allele at the position 10054 is 'CAA'
Screenshot 2021-03-25 at 15 58 44

However, the reference fasta points to the presence of allele 'CTA' instead.  

After fixing this discrepancy with bcftools norm, the resulting vcf looked like this. 

Screenshot 2021-03-25 at 16 09 01

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


jemten (approved):
👍

PR #1863, opened 2021-03-30
update vep version from 100 to 103

ramprasadn:
### This PR adds | fixes:

- See above

### How to test:

### Expected outcome:

### Review:
- [x] Code review
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ bwa-mem2 2.0 -> 2.2
 chanjo: 4.2.0 -> 4.6
 cyrius: v1.0 -> v1.1
 deepvariant: 1.0.0  
+gatk: v4.1.8.1 -> v4.2.0.0
 glnexus: v1.2.7
 htslib: 1.10.2-h78d89cc_0 -> 1.11-hd3b49d5_2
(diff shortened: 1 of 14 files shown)


jemten (commented):
There looks to be a 103.1 version out. Is that the same as 103?

ramprasadn:
> There looks to be a 103.1 version out. Is that the same as 103?

I noticed that, but docker image for 103.1 was uploaded before 103. That's why I went with 103 :D

jemten (approved):
I would go with 103.1 docker. If you look at their release tags on the github page, 103.1 is the more recent. https://github.com/Ensembl/ensembl-vep/tags The change seems very minor though :)
Otherwise it looks good 👍

ramprasadn:
> I would go with 103.1 docker. If you look at their release tags on the github page, 103.1 is the more recent. https://github.com/Ensembl/ensembl-vep/tags The change seems very minor though :)
> Otherwise it looks good 👍

I noticed that too. The naming for the dockers images are a bit misleading. There is 103.0 and 103.1, but they were both uploaded before 103. But I guess it makes sense to choose an image that corresponds to the latest github tag. Will go with 103.1 :)

PR #1886, opened 2021-04-14
feat(rna): resource bumps and fixes

jemten:
### This PR adds | fixes:

- Use the gencode annotation covering the primary assembly instead of only the chromosomes
- Scaled down the memory used for samtools sort
- RNA resource bumps

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/templates/mip_install_config.yaml
+++ b/templates/mip_install_config.yaml
@@ -117,7 +117,7 @@ container:
   mip:
     executable:
       mip:
-    uri: docker.io/clinicalgenomics/mip:9.1.3
+    uri: docker.io/clinicalgenomics/mip
   multiqc:
     executable:
       multiqc:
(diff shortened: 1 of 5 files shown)


templates/mip_install_config.yaml
```diff
@@ -117,7 +117,7 @@ container:
   mip:
     executable:
       mip:
-    uri: docker.io/clinicalgenomics/mip:9.1.3
+    uri: docker.io/clinicalgenomics/mip
```
henrikstranneheim, line 120: We just have to remeber to lock this when going to master later

henrikstranneheim (approved):
👍

templates/mip_install_config.yaml
jemten: Yes. QCcollect kept failing for me beacuse of a change we made in the cli in the dev branch. That's why I changed it

PR #1901, opened 2021-05-04
feat(bind_paths): Don't overwrite executable bind paths

jemten:
### This PR adds | fixes:
Previously the bind paths for an executable would be overwritten if the executable was used in more than one recipe. This appends the bind paths to the array rather than overwriting it.

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass


henrikstranneheim (approved):
👍

PR #1956, opened 2021-09-23
fixing annotation recipe

jemten:
### This PR adds | fixes:

- The loqus db headers weren't added to the correct file

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/definitions/rd_dna_initiation_map.yaml
+++ b/definitions/rd_dna_initiation_map.yaml
@@ -77,13 +77,13 @@ CHAIN_ALL:
     - rhocall_ar
     - bcftools_norm
     - variant_annotation
-    - mt_annotation
     - CHAIN_RHOVIZ:
       - rhocall_viz
       - chromograph_rhoviz
     - CHAIN_UPD:
       - upd_ar
       - chromograph_upd
+    - mt_annotation
     - frequency_filter
     - cadd_ar
     - varianteffectpredictor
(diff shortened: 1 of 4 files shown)


definitions/rd_dna_initiation_map.yaml
```diff
@@ -77,13 +77,13 @@ CHAIN_ALL:
...
     - bcftools_norm
     - variant_annotation
-    - mt_annotation
     - CHAIN_RHOVIZ:
       - rhocall_viz
       - chromograph_rhoviz
     - CHAIN_UPD:
       - upd_ar
       - chromograph_upd
+    - mt_annotation
```
projectoriented, line 86: Are the steps of the pipeline initiated from this yaml file from top down? What's the effect of mt_annot being moved here?

jemten: Shouldn't be any difference in the output of mt_annotation. However, upd_ar takes the output from the upstream recipe on the main chain which used to be the mt annotated file. While rhocall_viz is hardcoded to take the outfiles from variant_annotation. mt_annotation will always use the outfiles from variant _annotation regardless

projectoriented: Ahh okay, got it. 👍

projectoriented (approved):

jemten:
I'll merge this now. There has been so many dangling PR:s that I'm hard pressed to remember how everything fitted together 😅 . There's quite possibly gonna be a new one that tries to tie everything together

PR #1990, opened 2022-05-24
updates expansionhunter and saves output

### LINKED ISSUE #1982: Tagging ExHu files for export to HK and Scout reviewer service
**Is your feature request related to a problem? Please describe.**
Expansion Hunter generates some files that Scout Reviewer Service would need to generate images for Scout.
See https://github.com/Clinical-Genomics/project-planning/issues/361 for details.

Basically, sorted and indexed graph realignment bam-lets plus individual level ExHu output VCFs should do the trick. The sorting and indexing may not be done currently, so another set of samtools calls may be needed?

We still want the Stranger annotated, family VCFs for Scout though, so this is only a request to add file tagging, not change any current ones.

We don't have keys for this yet, so naming is free, but maybe something like `EXHU_BAMLET` and `EXHU_IND_VCF` on the sample/individual level.

The individual VCFs could potentially be recovered by splitting the family VCFs, but some filelds are somewhat convoluted so unless there are reasons against it, it would be easier just to run with the originals.

jemten:
### This PR adds | fixes:

- fixes #1982 

### How to test:

### Expected outcome:

### Review:
- [ ] Code review
- [ ] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Develop]
+
+- Save raw files from ExpansionHunter
+
+### Tools
+
+ExpansionHunter: 4.0.2 -> 5.0.0
+
 ## [11.0.1]
 
 - When running Deepvariant, set tmpdir to analysis folder and use `intermediate_results_dir`.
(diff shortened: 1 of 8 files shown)


jemten:
Need to sort the bam file also. Reverting to draft until fixed

ramprasadn (approved):
👍🏻

